PR #1, opened 2021-04-22
Update S3 policy for the EKS workers

scbarber:
Our site cloning feature relies upon the S3 API calls `ListObjectsV2` and `CopyObject` so we need those actions. In order to use `ListObjectsV2` we also need `ListBucket` so that has been added as well.


djmetzle:
CR :ear:

caphene:
CR 👍 

Good work, Seth!

zetas:
CR 🦖

PR #2, opened 2021-05-26
Add the Sentry configuration file

scbarber:
Our app code can consume a `/etc/dozuki/sentry.json` file to add a list of tags when sending exceptions to Sentry. This adds the `deployment` tag.


cloudprem/kubernetes.tf
```diff
@@ -109,31 +117,6 @@ resource "kubernetes_config_map" "dozuki_resources" {
     EOF
 
     "rds-ca.pem" = file(local.is_us_gov ? "vendor/rds-ca-${data.aws_region.current.name}-2017-root.pem" : "vendor/rds-ca-2019-root.pem")
-
```
zetas, line 112: did you intend to delete this index.json map? It should be added to, see the s3 or buckets config in this blob for the proper format.

scbarber: We aren't reading it in the app code any longer, so it should/can be gone.

zetas:
CR 👍

PR #3, opened 2021-06-09
Webhooks & Other Improvements

zetas:
# Background
Frontegg webhooks were added to the application and deployed on prod. While this solution would technically work in cloudprem, it would expose our customer's webhook data to a third party vendor. Usually our customers opt for cloudprem to keep tight control of their data and this would be the antithesis of that. 

# Solution
To support the data protection our clients expect we had to deploy additional infrastructure to effectively mimic frontegg's service in our individual sequestered cloudprem installations. They call this "hybrid mode". It requires considerable additional infrastructure including Apache Kafka, Redis, MongoDB, and additional MySQL databases, but provides total end-to-end control of the webhook lifecycle. I added a new module for our terraform deployment that creates the required infrastructure and wires it up to the application.

# QA
Deploy this stack using a custom config repo with the value `enable_webhooks` set to `true`. Also the replicated customer you are using need to have "Enable Webhooks" set to true in their customer edit screen in the vendor dashboard. All other frontegg config you see there should be left alone, the default values are correct and shared for all customers. 

# Notable Changes
Along with the webhook additions there are considerable other seamless improvements made to the terraform deployment:
1. Support for specifying existing S3 buckets and VPC. Previously these had to be imported to the terraform state manually.
2. S3 buckets now use versioning.
3. Replicated is now configured automatically to use a self-signed cert with the loadbalancer hostname and a default password for the dashboard is "dozuki". 
4. Part of the automated config is the ability to target a specific replicated sequence number for fresh installs using the variable `replicated_app_sequence_number`. A value of `0` will target the latest release in the channel assigned.
5. The Replicated scheduler has been updated to version 2.52
6. Various terraform providers have been updated to the latest versions. Most notably the AWS and Helm providers.
7. Infracost config and usage estimations have been added to the repo but not yet fully integrated into our Gitops. To run on your local machine, install Infracost and run `INFRACOST_TERRAFORM_BINARY=terragrunt infracost breakdown --usage-file infracost-usage.yml --path ./live/development` from the root of the repo.


live/development/terragrunt.hcl
```diff
@@ -10,7 +10,7 @@ terraform {
 
 # Configure input values for the specific environment being deployed:
 inputs = {
-  region = "us-west-2"
+  region = "us-east-1"
 
-  environment = "dev"
+  environment = "test"
```
djmetzle: Just to ask, this is setting the `environment` to `test` in the `development` folder. That seems a little off. Any reason to not fork off a `test` terragrunt folder?

cloudprem/modules/webhooks/versions.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  required_version = ">= 0.13.0"
+
+  required_providers {
+    aws  = ">= 3.2.0"
+    null = "~> 2.0"
```
djmetzle, line 6: :eyes:

cloudprem/modules/webhooks/variables.tf
```diff
@@ -0,0 +1,49 @@
...
+  description = "Name of the MSK cluster"
+}
+variable "rds_kms_key_id" {
+  type = string
+  default = ""
+  description = "KMS Key ID"
+}
+variable "cluster_size" {
+  type = number
+  default = 2
```
djmetzle, line 23: Is `2` a good default here? Do these guys build an "on-cluster" zookeeper? `2` is probably a bad value for Raft..

We're using MSK for managed Kafka? Or is this run "bare metal" in the cluster?

cloudprem/modules/webhooks/outputs.tf
```diff
@@ -0,0 +1,8 @@
+output "zookeeper_connect_string" {
+  value = aws_msk_cluster.this.zookeeper_connect_string
```
djmetzle: _Very_ interesting..

cloudprem/modules/webhooks/main.tf
```diff
@@ -0,0 +1,333 @@
...
+    value = "slack"
+  }
+  set {
+    name = "connectors-worker.frontegg.emails.provider"
+    value = "sendgrid"
+  }
+  set {
+    name = "connectors-worker.frontegg.emails.sendgrid.apiKey"
+    value = "dummyval"
+  }
```
djmetzle: Just to ask, this looks really boilerplatey. Any way we can use a `value.yml` with the helm provisioner? Just to help clean up all these individual `set` blocks? Or does the `set_sensitive` have to happen to guard secrets? Maybe we could come up a hybrid approach, of the "plaintext" values in their own file, but the sensitive values blocked out here?

Not a blocker, just asking...

cloudprem/modules/webhooks/charts/connectivity/templates/docker-registry-secret.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.frontegg.images.enabled }}
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/dockerconfigjson
+metadata:
+  name: regcred
+data:
+  .dockerconfigjson: {{ template "imagePullSecret" . }}
+{{- end }}
```
djmetzle, line 9: Ehw.. :eyes:

cloudprem/modules/webhooks/charts/connectivity/charts/integrations-service/templates/NOTES.txt
```diff
@@ -0,0 +1,2 @@
+1. Get the application URL by running these commands:
+
```
djmetzle, line 2: _Which_ commands? :sweat_smile:

cloudprem/modules/webhooks/charts/connectivity/charts/webhook-service/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v2
+name: webhook-service
+description: Frontegg Webhook Service Helm chart for Kubernetes
+type: application
```
djmetzle, line 4: Just to ask, this is basically a copy+paste of the upstream Frontegg Helm chart, including it's deps?

Maybe this gets updated in later commits, but we'll probably want some kind of README to document exactly how to update these charts. That's a notoriously ugly problem, merging charts. Very prone to conflicts. We'll want some breadcrumbs i think to the upsteam source for these charts?

zetas:
Aaaaand I just realized this is a public repo and I put the new default dashboard password in the PR text 🤦. I guess my next PR will be to use an autogenerated password for this instead and output it using terraform along with the dashboard url 🤦🤦🤦🤦🤦

cloudprem/modules/replicated/charts/replicated/templates/replicated.yaml
```diff
@@ -142,8 +144,9 @@ spec:
               mountPath: /var/run/replicated
             - name: docker-socket
               mountPath: /host/var/run/docker.sock
-            - name: replicated-conf
-              mountPath: /host/etc/replicated.conf
+#            - name: replicated-conf
+#              mountPath: /host/etc/replicated.conf
+#              subPath: replicated.conf
```
djmetzle: Should we leave these mounts commented? Seems a little crufty and misleading...

cloudprem/eks.tf
```diff
@@ -66,10 +66,7 @@ data "aws_iam_policy_document" "eks_worker" {
       "s3:PutObject",
       "s3:PutObjectAcl",
       "s3:GetObject",
-      "s3:GetObjectAcl",
-      "s3:ListBucket",
-      "s3:ListObjectsV2",
-      "s3:CopyObject"
```
djmetzle: Woah. I'm not sure i grok whats happening here. Does this `data` block fetch the AWS default policy for the `eks_workers`? If so, why do we need to specify actions for a `data` source? I'm a little confused by this...

cloudprem/eks.tf
```diff
@@ -118,7 +115,7 @@ module "eks_cluster" {
   source  = "terraform-aws-modules/eks/aws"
   version = "14.0.0"
 
-  depends_on = [module.vpc]
+  depends_on = [module.vpc, aws_iam_policy.cluster_access, aws_iam_policy.eks_worker]
```
djmetzle, line 121: Nice.

cloudprem/variables.tf
```diff
@@ -175,7 +175,13 @@ variable "cache_instance_type" {
   default     = "cache.t2.small"
 }
 
-# --- END Databsae and storage Options --- #
+variable "enable_webhooks" {
+  description = "This option will spin up a managed Kafka & Redis cluster to support private webhooks."
+  type        = bool
+  default     = false
+}
```
djmetzle, line 203: Might be outdated.. this variable to key the webhooks module isn't used quite yet?

cloudprem/main.tf
```diff
@@ -245,6 +245,26 @@ module "memcached" {
...
+  subnet_ids = local.private_subnet_ids
+
+  allowed_cidr_blocks = [local.vpc_cidr]
+
+  cluster_size  = local.azs_count
+  rds_kms_key_id = var.rds_kms_key_id
+
+  rds_address = module.primary_database.this_db_instance_address
+  rds_user = module.primary_database.this_db_instance_username
+  rds_pass = random_password.primary_database.result
```
djmetzle, line 314: We're using the "primary" RDS instance for webhooks as well? The description mentioned giving it it's own DB/RDS? Might be outdated...

cloudprem/static/userdata.yml
```diff
@@ -36,7 +36,15 @@ write_files:
 
     aws eks update-kubeconfig --name "${eks_cluster_name}" --role-arn "${eks_cluster_access_role}" --kubeconfig /home/ssm-user/.kube/config --region ${aws_region}
     chown ssm-user:ssm-user /home/ssm-user/.kube -R
+- path: /tmp/create_frontegg_schema.sh
+  permissions: '700'
+  content: |
+    #!/bin/bash -xe
+    set -o xtrace
+
+    mysql --defaults-file=/home/ssm-user/.my.cnf -e "create database frontegg_events; create database frontegg_webhooks"
```
djmetzle, line 45: Ehw. We're doing the `frontegg_events` DB creation during bastion boot? That's a little dirty. But, hey, if it works...

cloudprem/static/userdata.yml
```diff
@@ -36,7 +36,15 @@ write_files:
 
     aws eks update-kubeconfig --name "${eks_cluster_name}" --role-arn "${eks_cluster_access_role}" --kubeconfig /home/ssm-user/.kube/config --region ${aws_region}
     chown ssm-user:ssm-user /home/ssm-user/.kube -R
+- path: /tmp/create_frontegg_schema.sh
+  permissions: '700'
+  content: |
+    #!/bin/bash -xe
+    set -o xtrace
```
djmetzle, line 43: A little more canonical:
```suggestion
#!/bin/bash
set -exuo pipefail
```

cloudprem/database.tf
```diff
@@ -298,20 +288,23 @@ resource "aws_dms_replication_task" "this" {
...
-#   provisioner "local-exec" {
-#     when    = destroy
-#     command = "aws dms stop-replication-task --replication-task-arn ${self.triggers["dms_task_arn"]}"
-#   }
-# }
\ No newline at end of file
+ resource "null_resource" "replication_control" {
+   count = var.enable_bi ? 1 : 0
+
+   depends_on = [module.replicated]
```
djmetzle, line 294: Interested. This seems like graph hacking.

Maybe we could leave a comment here explaining the connection? Some like:
```
   # Once the replicated module is running, we know that DMS is ready to poke
```

Thoughts?

cloudprem/database.tf
```diff
@@ -273,19 +273,9 @@ resource "aws_dms_replication_task" "this" {
   replication_task_id      = local.identifier
   migration_type           = "full-load-and-cdc"
   replication_instance_arn = aws_dms_replication_instance.this[0].replication_instance_arn
-  table_mappings           = "{ \"rules\": [ { \"rule-type\": \"selection\", \"rule-id\": \"1\", \"rule-name\": \"1\", \"object-locator\": { \"schema-name\": \"%\", \"table-name\": \"%\" }, \"rule-action\": \"include\", \"filters\": [] }, { \"rule-type\": \"selection\", \"rule-id\": \"2\", \"rule-name\": \"2\", \"object-locator\": { \"schema-name\": \"mysql\", \"table-name\": \"%\" }, \"rule-action\": \"exclude\", \"filters\": [] }, { \"rule-type\": \"selection\", \"rule-id\": \"3\", \"rule-name\": \"3\", \"object-locator\": { \"schema-name\": \"performance_schema\", \"table-name\": \"%\" }, \"rule-action\": \"exclude\", \"filters\": [] } ] }"
+  table_mappings           = file("static/dms_mapping.json")
```
djmetzle, line 276: Much cleaner!

cloudprem/eks.tf
```diff
@@ -119,7 +119,7 @@ module "eks_cluster" {
 
   # EKS cofigurations
   cluster_name    = local.identifier
-  cluster_version = "1.18"
+  cluster_version = "1.19"
```
djmetzle, line 125: Sanity check...

Do we know that EKS v1.19 is available for use in all regions? Like govcloud?

cloudprem/kubernetes.tf
```diff
@@ -1,3 +1,11 @@
+data "kubernetes_secret" "frontegg" {
+  depends_on = [module.replicated]
+  count = var.enable_webhooks ? 1 : 0
+  metadata {
+    name = "frontegg-credentials"
+    namespace = "default"
+  }
+}
```
djmetzle, line 14: Oh, that's _cool_.

cloudprem/kubernetes.tf
```diff
@@ -72,10 +81,10 @@ resource "kubernetes_config_map" "dozuki_resources" {
...
       {
         "default": {
-          "guide-images": "${module.guide_images_s3_bucket[0].this_s3_bucket_id}",
-          "guide-pdfs": "${module.guide_pdfs_s3_bucket[0].this_s3_bucket_id}",
-          "documents": "${module.documents_s3_bucket[0].this_s3_bucket_id}",
-          "guide-objects": "${module.guide_objects_s3_bucket[0].this_s3_bucket_id}"
+          "guide-images": "${local.guide_images_bucket}",
+          "guide-pdfs": "${local.guide_pdfs_bucket}",
+          "documents": "${local.documents_bucket}",
+          "guide-objects": "${local.guide_objects_bucket}"
```
djmetzle, line 100: I don't see where we're specifying these locals. Maybe we could split these resources out into their own file? Just to get a little better locality?

cloudprem/main.tf
```diff
@@ -33,21 +33,27 @@ locals {
...
+  private_subnet_ids = local.create_vpc ? module.vpc[0].private_subnets : data.aws_subnet_ids.private
 
   # Database
   ca_cert_identifier = local.is_us_gov ? "rds-ca-2017" : "rds-ca-2019"
 
+  # S3 Buckets
+  guide_images_bucket = var.create_s3_buckets ? module.guide_images_s3_bucket.this_s3_bucket_id : data.aws_s3_bucket.guide_images[0].bucket
+  guide_objects_bucket = var.create_s3_buckets ? module.guide_objects_s3_bucket.this_s3_bucket_id : data.aws_s3_bucket.guide_objects[0].bucket
+  guide_pdfs_bucket = var.create_s3_buckets ? module.guide_pdfs_s3_bucket.this_s3_bucket_id : data.aws_s3_bucket.guide_pdfs[0].bucket
+  documents_bucket = var.create_s3_buckets ? module.documents_s3_bucket.this_s3_bucket_id : data.aws_s3_bucket.documents[0].bucket
```
djmetzle, line 55: There they are!

cloudprem/main.tf
```diff
@@ -60,6 +66,11 @@ data "aws_availability_zones" "available" {}
 
 data "aws_caller_identity" "current" {}
 
+data "aws_vpc" "this" {
+  count = local.create_vpc ? 0 : 1
+  id = var.vpc_id
+}
```
djmetzle, line 72: Odd. A `count` specified in a `data` block? I'm surprised that even works...
It's a little odd that we're "flipping" the count, so `count = 0` for `create_vpc = true`. I think the naming here might be confusing. Calling it `this` isn't particularly descriptive. Maybe something like `existing_vpc` or the like could be helpful?

cloudprem/main.tf
```diff
@@ -33,21 +33,27 @@ locals {
     Environment = var.environment
   }
 
-  protect_resources = false #var.stack_type == "prod" ? true : false
+  protect_resources = var.stack_type == "production" ? true : false
```
djmetzle, line 36: What do you think about splitting this out into it's own variable? It would make things a little more explicit i think. We could have terragrunt breakout `protect_resources` in the defaults for an environment? That way we can toggle it intentional from the config side, and it's not implicit on the stack type. I can totally imagine a case where we'd want to QA the "production mode", but still be able to easily destroy that stack when we're done. Thoughts?

cloudprem/outputs.tf
```diff
@@ -5,10 +5,10 @@ output "eks_cluster_access_role" {
 
 output "dashboard_url" {
   description = "URL to your Dozuki Dashboard."
-  value       = "${module.nlb.this_lb_dns_name}:8800"
+  value       = format("https://%s:8800",module.nlb.this_lb_dns_name)
```
djmetzle, line 8: Does `terraform fmt` not balk at the lack of whitespace here? Might be outdated....

cloudprem/variables.tf
```diff
@@ -233,4 +254,30 @@ variable "environment" {
...
     error_message = "The length of the Environment must be less than 6 characters."
   }
+}
+variable "stack_type" {
+  description = "Specify whether this is a production stack or a development stack. Production stacks have data protection settings enabled but will leave resources behind after a terraform delete operation."
+  type = string
+  default = "production"
+
+  validation {
+    condition = var.stack_type == "dev" || var.stack_type == "production"
```
djmetzle, line 264: Just a quick thought... maybe we can use a [terraform function](https://www.terraform.io/docs/language/functions/contains.html) to check against a list of good values? That could help make this rule a little more declarative for maintenance?

cloudprem/variables.tf
```diff
@@ -233,4 +254,30 @@ variable "environment" {
...
+variable "frontegg_client_id" {
+  description = "If webhooks are enabled, this is the client_id for access to Frontegg's servers."
+  default = ""
+  type = string
+}
+
+variable "replicated_app_sequence_number" {
+  description = "For fresh installs you can target a specific Replicated sequence for first install. This will not be respected for existing installations"
+  default = "latest"
+  type = string
```
djmetzle: Could we get away with an `int` here, and `toString` it where it's used? The sequence numbers are always int valued, right?

cloudprem/modules/replicated/replicated.json
```diff
@@ -0,0 +1,7 @@
+{
+  "DaemonAuthenticationType": "password",
+  "ReleaseSequence": ${release_sequence},
+  "DaemonAuthenticationPassword": "dozuki",
+  "TlsBootstrapType": "self-signed",
+  "TlsBootstrapHostname": "${nlb_hostname}"
```
djmetzle: Hmm. For consistency here, i think we either want to quote this ReleaseSequence expansion, or _unquote_ the TlsBootStrapHostname? I think i'd prefer to see the latter...

cloudprem/modules/replicated/variables.tf
```diff
@@ -5,4 +5,8 @@ variable "dozuki_license_parameter_name" {
 variable "nlb_hostname" {
   description = "The hostname of the deployed network loadbalancer"
   type = string
+}
+variable "release_sequence" {
+  description = "Pin a replicated release sequence on firstboot."
+  default = ""
```
djmetzle, line 11: Hmm. If we're defaulting this in the calling module, maybe we should set a default here in the inner module?

cloudprem/modules/webhooks/main.tf
```diff
@@ -102,76 +92,41 @@ resource "aws_msk_cluster" "this" {
...
-  name       = "${var.name}-mongo-subnet_group"
-  subnet_ids = var.subnet_ids
 
-  tags = var.tags
-}
-resource "aws_docdb_cluster_parameter_group" "disable_tls" {
-  family      = "docdb4.0"
-  name        = "${var.name}-mongo-param-grop"
-  description = "docdb cluster parameter group"
+resource "helm_release" "mongodb" {
```
djmetzle, line 98: This pulls mongo into the cluster as well?

cloudprem/modules/webhooks/main.tf
```diff
@@ -102,76 +92,41 @@ resource "aws_msk_cluster" "this" {
...
+resource "helm_release" "redis" {
 
-resource "aws_elasticache_subnet_group" "default" {
-  name       = "${var.name}-redis-subnet-group"
-  subnet_ids = var.subnet_ids
-}
+  name = "frontegg-kvstore"
+  repository = "https://charts.bitnami.com/bitnami"
+  chart = "redis"
+  version = "14.1.1"
```
djmetzle, line 116: Did frontegg struggle with managed redis? We're pull a _lot_ of storage and state into the cluster, which could be problematic. I suspect "small" (read: dev/QA) clusters might struggle to run kafka+mongo+redis?

I think we should prefer to use managed services whereever possible.. but i suspect it just didn't work with frontegg?

cloudprem/modules/replicated/charts/replicated/templates/replicated.yaml
```diff
@@ -77,9 +77,11 @@ spec:
         app: replicated
         tier: master
     spec:
+      imagePullSecrets:
+        - name: replicateddockerhubkey
```
djmetzle, line 81: It's lame that we need that... Just to pull the replicated container images too?

cloudprem/main.tf
```diff
@@ -314,7 +314,6 @@ module "webhooks" {
   rds_pass = random_password.primary_database.result
 
   eks_sg = module.eks_cluster.cluster_primary_security_group_id
-  eks_cluster = module.eks_cluster
```
djmetzle: This is a little surprising. Doesnt the frontegg module spray a bunch of helm charts into the cluster? I don't see how it would be able to connect?

djmetzle (approved):
Awesome work! I've put together a lot of comments, but many are outdated. I'll try to resolve the ones that were fixed in later commits in this PR. Again, awesome work here @zetas!

CR :gear:

live/development/terragrunt.hcl
djmetzle: Fixed in a later commit: https://github.com/Dozuki/CloudPrem-Infra/pull/3/commits/63531b68d368347d8a99db346988bdc811b2da7f

cloudprem/modules/webhooks/variables.tf
zetas: That variable value will actually never be used so it should be removed. The cluster size minimum for MSK is equal to the amount of AZ's in the VPC it's deployed into so that's what I set it to. In the future if that's not enough I'll expose the variable to the customer config so they can increase it.

cloudprem/modules/webhooks/main.tf
zetas: Yes, frontegg requires mongo so if webhooks are enabled a small standalone mongo server is installed on cluster.

cloudprem/modules/webhooks/charts/connectivity/charts/webhook-service/Chart.yaml
djmetzle: Yeah, there's a later commit that does just this: https://github.com/Dozuki/CloudPrem-Infra/pull/3/commits/4d15885adf5d20903770f50203fc55c3653513e4

It might be helpful to have a README on _how_ to update the charts. Even if it's just some links off to where the upstream charts (currently!) live?

cloudprem/modules/replicated/charts/replicated/templates/replicated.yaml
djmetzle: Removed later..

cloudprem/eks.tf
djmetzle: Yep. These got put right back in: https://github.com/Dozuki/CloudPrem-Infra/pull/3/commits/0bcf116cda2e91d0223998e4eadf75724eec3b9e

cloudprem/variables.tf
djmetzle: Used later...

cloudprem/modules/webhooks/main.tf
zetas: Actually we would prefer not to use managed services for things like this. It's such a small use and requires no clustering or advanced features. Their use of redis is basically a queue store so there's no persistent data stored there so there is little risk of data loss. I didn't see a reason to add $20 or so to the monthly bill for no real gain but it did work fine with elasticache redis so if we wanted to use that instead we could. It did not, however, work with aws's mongo (documentDB). That requires a pure mongo server.

cloudprem/modules/replicated/charts/replicated/templates/replicated.yaml
zetas: IDK if we "need" that. This is the yaml that was generated by the update replicated release shell script provided by them. Something must've changed since the last version.

cloudprem/main.tf
zetas: The helm provider implicitly inherits the eke cluster context from the calling module so no direct connection is required.

cloudprem/modules/webhooks/main.tf
zetas: Yup, it's now in a values.yaml :)

cloudprem/modules/webhooks/charts/connectivity/charts/webhook-service/Chart.yaml
zetas: Sure, good call, we should include the chart update process in the readme. And yea, it's literally just a copy of their charts.

cloudprem/main.tf
zetas: I didn't see a reason to spin up an additional RDS database just for this so we create the frontegg schemas on the main DB. Open to suggestions if that's problematic.

cloudprem/static/userdata.yml
zetas: It's super dirty and I hate it. But short of modifying the frontegg helm charts to add the support for init containers this was the best way I could think of doing it that would work both on a local laptop as well as in the pipeline (terraform has a mysql resource that could do it but that would not work on a laptop due to inaccessibility of the RDS instance from outside the VPC)

cloudprem/database.tf
zetas: This _is_ kind of graph hacking but it's not as bad as you might think. I added this dependency because I don't want DMS to start until the application is deployed and there is some data to replicate. I saw some issues where DMS was unable to pickup new schemas after the replication task was started. It's a bug in DMS but this is an attempt to prevent it from happening. It's effectively baking a race condition in but without visibility into replicated's status wrt the application deployment there's not really any other way to achieve it. Open to suggestions tho :)

cloudprem/eks.tf
zetas: I did not check "all" regions but it definitely works in govcloud us-gov-east-1 (which is the only govcloud region we support anyway due to the lack of code pipeline in west)

cloudprem/kubernetes.tf
zetas: The local definitions are in the main.tf at the top. This whole thing is undesirable from an organization perspective so any recommendations are welcome.

cloudprem/main.tf
zetas: It's the only VPC which is why I use the "this" title, though we could definitely rename it. The reason we use count is because that data block should be invoked when an existing vpc is specified. If an existing VPC isn't specified and that data block was not conditional it would generate an error and fail the run (VPC of blank can't be found lol). Basically saying: if vpc exists, grab it's info, otherwise grab the info from the created vpc.

cloudprem/modules/webhooks/main.tf
djmetzle: Nice! :+1:

cloudprem/outputs.tf
zetas: Lack of whitespace where exactly? After the parameter comma?

cloudprem/main.tf
djmetzle: No, that's fine. It just makes the PR description a little misleading..

I dont have any problem with frontegg using the main RDS... Might be something to keep an eye on though performance-wise?

cloudprem/variables.tf
zetas: I'll definitely look into that. This variable definition was created by our contractors.

cloudprem/modules/replicated/replicated.json
zetas: Can't do that because TlsBootStrapHostname NEEDS to be a string. ReleaseSequence NEEDS to be an integer. If you pass an integer in quotes replicated will ignore it 🙄

cloudprem/static/userdata.yml
djmetzle: Oh. Ehw. The frontegg chart doesn't allow init containers. That explains it.

cloudprem/modules/replicated/variables.tf
zetas: Yea for sure, I'll pull that out.

cloudprem/database.tf
djmetzle: Oh, i see. That makes more sense. But does the `replicated` module actually wait for the app to spin up?

cloudprem/kubernetes.tf
djmetzle: Hmm. We could split out a top-level file for the configmap resources. But that doesn't really help locality. Though it might be a nice organization tweak. Maybe we can split out the S3 resources into a separate file with it?
https://github.com/Dozuki/CloudPrem-Infra/blob/c90fea4d76dfa60f5a3cca8bc8c7cccc59fb5b9c/cloudprem/main.tf#L121-L145
Or the configmap?
https://github.com/Dozuki/CloudPrem-Infra/blob/c90fea4d76dfa60f5a3cca8bc8c7cccc59fb5b9c/cloudprem/kubernetes.tf#L29-L161

That configmap resource could totally be in a file by itself.

I dont really know though. I dont think TF makes locality or organization very easy. This is probably fine for now...

cloudprem/main.tf
djmetzle: I see.

Might it better to just parameterize then? We could pass in `create_vpc<bool>` and `existing_vpc<?vpcid>` from the config? Otherwise, we're doing some implicit conditional trickery?

cloudprem/outputs.tf
djmetzle: Yep! `terraform fmt` doesn't complain about that?

cloudprem/modules/replicated/replicated.json
djmetzle: Gotcha.

djmetzle: JSON building specific.

cloudprem/modules/webhooks/main.tf
djmetzle: All the storage services!

cloudprem/modules/webhooks/main.tf
djmetzle: Okay sure. My concern is more around running storage in cluster. So if these services start thrashing the cluster, they could effect the app critical path.

Also, i think we could see surprising behavior as we replace nodes. If we replace the whole node ASG, those pods are going to have to move around, _AND_ they're stateful, so they'll be down for quite a bit while the PVs move around. 

I think that's worth pointing out. It might allow for some dataloss of inflight webhook events during scaling..

cloudprem/modules/replicated/charts/replicated/templates/replicated.yaml
djmetzle: Alright. If that's auto-generated, lets not worry about it. :+1:

cloudprem/main.tf
djmetzle: Ah! So the helm provider picks up the connection and inherits? Nice.

cloudprem/database.tf
zetas: No it doesn't. Hence the baked in race condition. It's better than nothing but it is certainly not deterministic. The helm chart is only concerned with "is the replicated scheduler installed successfully". The app could explode and it would still succeed.

djmetzle: Oh. I see. That's no good.

Here's an idea, and a filthy hack...

What do you think about using a `local-exec` or the likes to poll the DB for the db-initialize to at least start? We could spinlock looking for the `onprem_guide` schema to exist. That way we'll know DMS has _something_ to copy?

zetas: The problem with that is terraform will not have access to the mysql database from a laptop and I really don't want to add any terraform code that locks us out of deploying these stacks on our workstations if possible. Same reason why we have to use that bastion user data hack to create the frontegg schemas. The "correct" answer is probably to create a lambda that will actually poll the app endpoint and hold the DMS start until it shows up but that's a lot more work for a relatively minor issue. After all if the BI database is not populated then the customer will know that and can just restart the replication task manually.

djmetzle: Alright. So as long as this fails in an acceptable way under the worst case, then it's probably fine. This is probably "close enough" for most updates?

scbarber (approved):
CR :barber: LGTM

cloudprem/modules/webhooks/versions.tf
caphene: wat

caphene (commented):
CR :+1: 

Had one small question but just for my own clarification. :)

PR #4, opened 2021-09-09
Add memcached test cluster configs

bcooper94:
Background
--
This configuration is used to determine how to connect to the memcached server that's used to acquire mutexes for nanosites when running integration tests. This may not get us all the way to being able to run all integration tests in CloudPrem, but it will at least get us one step closer.

QA
--
I'm not sure how QA is being done on CloudPrem these days, but this should have no effect on the application's regular behavior.

CC @scbarber @ddvdozuki


scbarber (approved):
CR 💈

djmetzle (approved):
CR :test_tube:

PR #5, opened 2021-09-13
Bastion userdata fix

ddvdozuki:
I somehow messed up the formatting on the user data embedded script which broke the user data entirely so there's no way to debug current cloudprem installations so this is classified as a hot fix.


djmetzle (approved):
CR :busts_in_silhouette:

caphene:
CR 🎸

PR #6, opened 2021-09-18
Major Terraform Refactor

ddvdozuki:
# Background
The main issue we had with our terraform code for cloudprem was reliability. Our stack spin-up and spin-down was failing ~50% of the time due to race conditions and expired tokens with Kubernetes/helm and changes being made outside terraform with null resource provisioners and misconfigurations in our EKS cluster module. These issues prevented our stack from being production ready.

# Solution
Following best practice to split our infrastructure into discreet modules with dependencies we now have 4 separate modules:
1. network
2. compute
3. storage
4. app

The dependency configuration is set so they execute sequentially in order with terragrunt. This gives us all kinds of benefits when managing stacks, most notably change isolation and module protection. If there's an issue with a module it doesn't necessarily affect the entire stack and you can delete and recreate each module independently with some caveats around dependencies. Read [here](https://blog.gruntwork.io/5-lessons-learned-from-writing-over-300-000-lines-of-infrastructure-code-36ba7fadeac1) for more info on split stack vs monolithic.

This also solves our race condition problems by separating our EKS deploy with our Kubernetes payload deploy, EKS is deployed in the compute module while the payload Is deployed in the app module. Running it this way prevents Just-in-time and interpolation issues with Kubernetes authentication and token management, see [here](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#stacking-with-managed-kubernetes-cluster-resources) for a warning from terraform directly

# QA
The live folder structure has changed to support the multi module setup and add some new features as well. The live testing system now consists of 3 levels of folders to organize the configurations by account, region, and environment. Additionally there are 4 environments stubbed out in each region: min, bi, webhooks, and full. It should be self explanatory what each of these environments is for but for more info checkout the README.md file inside the live folder.

To spin up this new terraform code for QA you would pick a region and an account first, let's say standard account and us-east-1 region, then follow the steps below:

1. Modify the `live/standard/account.hcl` file and input the AWS account ID you want to use. We use this variable to lock down the terraform deployment from accessing any other accounts you may have configured accidentally on your workstation. (do not commit these changes back to the repo)
2. cd into the environment you want to test, let's say `live/standard/us-east-1/min` for this example.
3. Run terragrunt to apply the initial stack (plan isn't supported for a new stack): `terragrunt run-all apply` and type `y` to confirm.
4. Go get a coffee

Once complete you'll be presented with 3 outputs: the replicated dashboard URL, the dashboard password, and the application url. Visit the app url and the dozuki login page should be present (you may need to wait 5 minutes or so after a fresh deploy for the app to finish starting up, if it takes longer than that login to the dashboard and investigate)

# Notable Changes
Along with the main terragrunt refactor I've updated the GitHub action workflow to now run TFSec, TFlint, and terraform format check. The old action workflow had these checks disabled for some reason. Additionally the TFSec and TFlint check will comment on the PR with any issues it detects.

Additionally I added pre-commit hook configuration to auto-run terraform docs in the modules on new commits. To use this with your local clone install pre-commit (with home-brew it's `brew install pre-commit`) and then run `pre-commit install` and `pre-commit install-hooks` in the repo root. If you commit code and get a failure on the pre-commit that's fine, that only means the README(s) have been updated.

# Caveats
This refactor is incompatible with our existing code pipeline and CloudFormation systems and will need some minor changes to maintain a similar workflow. For that reason even if/when this PR is merged will not be able to "release" this refactor until those changes are done as well. Thankfully we use Gitops for our terraform versioning so merging this into master won't break any existing installs.

Closes https://github.com/iFixit/dozuki/issues/2823
Closes https://github.com/iFixit/dozuki/issues/2822
Closes https://github.com/iFixit/dozuki/issues/2821
Closes https://github.com/iFixit/dozuki/issues/2820


cloudprem/storage/rds.tf
```diff
@@ -86,18 +84,19 @@ module "primary_database" {
 resource "aws_secretsmanager_secret" "primary_database_credentials" {
```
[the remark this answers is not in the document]
ddvdozuki: Resolved by adding TFSec exception. We should use CMK but this will require additional IAM policies and I didn't want to add them in this huge PR.

scbarber (approved):
CR :barber:

cloudprem/app/charts/kubed/.helmignore
```diff
@@ -14,10 +14,10 @@
 *.swp
 *.bak
 *.tmp
-*.orig
 *~
 # Various IDEs
 .project
 .idea/
 *.tmproj
-.vscode/
```
caphene, line 23: Why is this being removed? Same for *.orig on L17.

ddvdozuki: This appears to be GitHub getting confused. It thinks that the .helmignore from the frontegg integrations-service chart was replaced by the helmignore for the kubed chart. I'm not sure exactly what's goin on here but these files are maintained by the chart maintainers and are not like gitignore files. It basically tells helm to ignore certain files when processing the templates. With the frontegg chart I had to do some creative terraform to insert a helm ignore file at runtime with terraform but with the kubed chart I didn't have to so this is the original file.

cloudprem/storage/rds.tf
```diff
@@ -83,21 +81,23 @@ module "primary_database" {
   tags = local.tags
 }
 
+#tfsec:ignore:aws-ssm-secret-use-customer-key
 resource "aws_secretsmanager_secret" "primary_database_credentials" {
   name = "${local.identifier}-database"
 
-  recovery_window_in_days = local.protect_resources ? 7 : 0
+  recovery_window_in_days = 0
```
caphene, line 88: Is this intentional?

caphene (approved):
CR 👍 

Just had one question to double check.

PR #7, opened 2021-10-06
Add cluster/pod scaling and auto-healing for EKS

ddvdozuki:
# Background
Our cloudprem infra has always been lacking in auto-scaling, node termination handling, and other HA features built into EKS and Kubernetes. We've been able to get the app to work fairly well but we always left quite a bit on the table as far as auto-healing and redundancy is concerned and our infrastructure always had to be over-provisioned to make up for this.

# Solution
We will use the cluster autoscaler, node termination handler, horizontal pod autoscalers, and metrics collection along with modifying our core app configuration regarding replicas and resource requests to achieve a truly cloud-native, auto scaling, auto healing infrastructure. This has many obvious wins but perhaps the biggest one is our ability to leverage AWS spot fleets for our worker nodes. EC2 bills for this infra will now be considerably less (up to 90%) than our previous on-demand only setup. This also means we can bring nodes in and out as needed and react to all manner of workflows and workloads. 

# QA
Spin up a stack using the live configuration (see the `live/README.md` file for instructions) and follow these steps:

1. Configure your workstation to connect to the Kubernetes cluster like so:
  a. `aws eks update-kubeconfig --name dozuki-ENVNAME`
    1. Replace `ENVNAME` with whatever you set your `env` variable to in the terraform config. 
2. Verify the HPAs have been created:
  a. `kubectl get hpa -A`
    1. which should look like:
  ```
NAMESPACE                                     NAME               REFERENCE                            TARGETS           MINPODS   MAXPODS   REPLICAS   AGE
replicated-6c5c501278154a1a434794cf9ba68977   app-hpa            Deployment/app-deployment            45%/80%, 1%/50%   2         50        2          110m
replicated-6c5c501278154a1a434794cf9ba68977   queueworkerd-hpa   Deployment/queueworkerd-deployment   20%/80%, 1%/80%   2         50        2          110m
```
3. Verify the cluster autoscaler and node termination handler has been installed:
  a. `kubectl get po -n kube-system` 
    1. Example output:
 ```
NAME                                                         READY   STATUS    RESTARTS   AGE
...
aws-node-termination-handler-c9b74c595-dg9v9                 1/1     Running   0          131m
cluster-autoscaler-aws-cluster-autoscaler-7cd6c7cc88-l6xpj   1/1     Running   0          131m
...
```
4. Open the app, create a guide and upload a bunch of pictures to the media manager, the more the better. Come back to the console and watch the queueworkerd pod scale up after a few seconds by running `kubectl get po -A` look for something like this:
```
replicated-6c5c501278154a1a434794cf9ba68977   queueworkerd-deployment-6f7b4b8578-cm547                     1/1     Running   0          112m
replicated-6c5c501278154a1a434794cf9ba68977   queueworkerd-deployment-6f7b4b8578-ppbt6                     1/1     Running   0          112m
```
It will start with 2 but should grow to many more after a few seconds of processing

# Notable Changes
As I said above we are now using spot fleets instead of on-demand instances. Because of this I chose to use a larger default instance size. The suggested sizes are `["m5.large", "m5a.large", "m5d.large", "m5ad.large"]`. The way spot fleets work it will spin up nodes from any of those sizes but even though they are larger they should be quite a bit cheaper than the t3.medium on-demand nodes we were using before and because they are more powerful there are fewer scaling events needed (scaling the cluster is fine but it's not nearly as instant as scaling the pods so pod scaling should be preferable)


cloudprem/app/scaling.tf
```diff
@@ -0,0 +1,87 @@
...
+  depends_on = [helm_release.replicated]
+  metadata {
+    name = "app-hpa"
+    # Terraform magic required to find the randomly created replicated namespace so we can install the HPAs in the right place.
+    namespace = coalesce([for i, v in data.kubernetes_all_namespaces.allns.namespaces : try(regexall("replicated\\-.*", v)[0], "")]...)
+  }
+
+  spec {
+    min_replicas = 2
+    max_replicas = 50
```
djmetzle, line 15: I'd recommend starting with a lower max here. Otherwise a big scale out could swap the whole cluster.
```suggestion
    max_replicas = 10
```

cloudprem/compute/scaling.tf
```diff
@@ -0,0 +1,85 @@
+resource "aws_iam_policy" "cluster_autoscaler" {
```
djmetzle, line 1: There's at least 3 different resources of different kinds with the same name `cluster_autoscaler` here. That can be a little confusing in state and plan output. Maybe we can name these something like `cluster_as_policy`?

cloudprem/compute/scaling.tf
```diff
@@ -0,0 +1,85 @@
...
+
+    condition {
+      test     = "StringEquals"
+      variable = "autoscaling:ResourceTag/k8s.io/cluster-autoscaler/enabled"
+      values   = ["true"]
+    }
+  }
+}
+
+resource "helm_release" "cluster_autoscaler" {
```
djmetzle, line 50: We're installing the cluster autoscaler as an operator with helm? :thinking: 

I would think that'd be an EKS feature...

cloudprem/app/scaling.tf
djmetzle: The storage tier wont scale elastically though? So we need to be careful where we set the max thresholds.

PR #9, opened 2021-10-11
Adding Machine Translation config

ddvdozuki:
To support google machine translation in cloudprem we needed to add the supporting config values with the google translate api token. This adds that as well as the supporting variable for customers to add to their config.


scbarber (approved):
CR :barber:

PR #12, opened 2022-01-08
Terraform Refactor to Physical & Logical Modules

ddvdozuki:
# Background
What we're seeing with the 4 module approach is greater flexibility for development, being able to spin up and down just a small subset of our terraform but when It comes to end-user code pipeline use it creates quite a bit of added complexity, especially when handling partial creation failures. In the old stack if you started to spin up an environment and it failed on the storage or compute module but network was already successfully created, there was no way to delete the stack without manual workstation intervention because Terragrunt would always try to destroy the stack in reverse order (app, storage, compute, network) and if those modules weren't created yet, Terragrunt could not handle that and would just die. Additionally with 4 modules you had 4 greater opportunities at failure in both directions: create and destroy. 

# Solution
It turns out a modular approach was a good idea but it makes more sense to have fewer modules that bifurcate using different logic. We now have 2 modules: Physical - which includes all physical resources, like the k8 cluster, databases, ec2 instances, IAM roles & policies, etc, and Logical - which is all the resources that are deployed _onto_ the physical resources. So k8 manifests & helm charts mostly. Not only does this modular configuration reduce our error possibility but it makes environment deletion a snap. We only have to run the destroy command on the physical module since deleting the physical stuff will automatically delete all the logical resources deployed onto them. This makes for a much more reliable code pipeline as well. If something breaks somewhere, the deletion phase is only 1 module long so it will always be able to clean up after itself regardless of what might fail during creation. 

# QA
The live folder structure has changed to support the dual module setup. It should be self explanatory what each of these environments is for but for more info checkout the README.md file inside the live folder.

To spin up this new terraform code for QA you would pick a region and an account first, let's say standard account and us-east-1 region, then follow the steps below:

1. If using our testing AWS account and the default local awscli profile, you don't need to modify the account.hcl file anymore.
2. cd into the environment you want to test, let's say `live/standard/us-east-1/min` for this example.
3. Run terragrunt to apply the initial stack (plan isn't supported for a new stack): `terragrunt run-all apply` and type `y` to confirm.
4. Go get a coffee

Once complete you'll be presented with 3 outputs: the replicated dashboard URL, the dashboard password, and the application url. Visit the app url and the dozuki login page should be present (you may need to wait 5 minutes or so after a fresh deploy for the app to finish starting up, if it takes longer than that login to the dashboard and investigate)

# Notable Changes
Along with the main terragrunt refactor I've made many upgrades to the system as a whole:

1. Refactored S3 buckets to no longer use the terraform s3 module
   * This module was causing all kinds of issues with tfsec and honestly was way more complicated than we needed for simple buckets.
   * S3 buckets no longer allow public ACL's of any kind to follow AWS S3 best security practices.
   * We now make use of a logging bucket to log all S3 changes in an environment.
   * [Source for above](https://github.com/Dozuki/CloudPrem-Infra/blob/physical-logical/cloudprem/physical/s3.tf)
2.  Massive refactor of the DMS system for BI environments.
    * We wait until the replicated app is fully available before starting the DMS replication task. This ensures our entire database gets replicated on first start. For some reason DMS seems flakey when replicating database and table additions with our system. 
      * [source](https://github.com/Dozuki/CloudPrem-Infra/blob/physical-logical/cloudprem/logical/bi.tf)
    * We now use a utility shell script to stop the DMS task on destroy operations. Using the built-in null resource local-exec provisioner alone was not sufficient for this asynchronous process. The script now stops the task and waits for it to be fully stopped before allowing deletion to prevent errors 
      * [invocation](https://github.com/Dozuki/CloudPrem-Infra/blob/5f57a82e73ab986e6efc12606c66a4f6bff6f8db/cloudprem/physical/rds.tf#L331) and [source](https://github.com/Dozuki/CloudPrem-Infra/blob/physical-logical/cloudprem/physical/util/dms-stop.sh)
    * DMS now uses in-transit SSL for all communication between master and replica RDS instances. 
      * [source](https://github.com/Dozuki/CloudPrem-Infra/blob/5f57a82e73ab986e6efc12606c66a4f6bff6f8db/cloudprem/physical/rds.tf#L249)
    * Expanded DMS config to include data validation and error logging to CloudWatch. 
      * [source](https://github.com/Dozuki/CloudPrem-Infra/blob/physical-logical/cloudprem/physical/static/dms_config.json)
 3. All K8 secrets are now encrypted using KMS 
    * [source](https://github.com/Dozuki/CloudPrem-Infra/blob/5f57a82e73ab986e6efc12606c66a4f6bff6f8db/cloudprem/physical/eks.tf#L205)
 4. Upgraded many terraform modules and the AWS provider to the latest version available at the time.
 5. Linked the dashboard password resource to the NLB id. This forces a new password to be generated for the environment if physical-only deletion and full-stack recreation occurs. 
    * [source](https://github.com/Dozuki/CloudPrem-Infra/blob/5f57a82e73ab986e6efc12606c66a4f6bff6f8db/cloudprem/logical/replicated.tf#L9)

# Caveats
This refactor is incompatible with our existing code pipeline and CloudFormation systems and will need some minor changes to maintain a similar workflow. For that reason even if/when this PR is merged will not be able to "release" this refactor until those changes are done as well. Thankfully we use Gitops for our terraform versioning so merging this into master won't break any existing installs.


live/gov/account.hcl
```diff
@@ -1,6 +1,6 @@
 # Set account-wide variables. These are automatically pulled in to configure the remote state bucket in the root
 # terragrunt.hcl configuration.
 locals {
-  aws_account_id = "<ENTER AWS GOVCLOUD ACCT #>"
+  aws_account_id = "446787640263"
```
scbarber, line 4: Is this an intended change?

scbarber (commented):
Overall it looks good. I did have one question regarding a change which introduced an account ID....

live/gov/account.hcl
ddvdozuki: Yes, remember we discussed the fact that our standard and govcloud AWS ID was already available publicly (necessarily for access to our Lambda App) in our CloudFormation so it didn't make a lot of sense to censor it here. Including it for our testing platform adds a lot of convenience.

scbarber (approved):
CR 💈

live/gov/account.hcl
caphene: It's not entirely clear to me how valuable an AWS account ID is. It does seem like this lowers the barrier to collecting information compared to other locations. An example that comes to mind is a search like `aws_account_id site:github.com`.

I don't know enough about AWS account security to make a call either way on this but wanted to raise it as a potential concern just in case.

ddvdozuki: My point it is doesn't really matter since we have a hard requirement to have the AWS account ID public for our Lambda's to work, see here: https://github.com/Dozuki/CloudPrem/blob/add32e387eb41f69e1524c6753d835574e601fcc/codepipeline_app.yml#L59

physlog/physical/util/dms-stop.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
djmetzle: `set -euo pipefail`?

cloudprem/physical/util/dms-stop.sh
```diff
@@ -4,25 +4,35 @@ TRIGGER="$1"
...
-      STATUS=$(checkForStoppedDMS)
-   done
+  STATUS=$(getDMSStatus)
+  while [[ "$STATUS" == "stopping" ]]; do
+    sleep 10
+    STATUS=$(getDMSStatus)
+  done
 
-   echo -e "DMS Task Stopped Successfully."
+  echo -e "DMS Task Stopped Successfully."
```
djmetzle, line 35: Do you want to double check that `STATUS == 'stopped'`? Is there any other states it could be here?

cloudprem/physical/eks.tf
```diff
@@ -194,7 +194,8 @@ module "eks_cluster" {
   cluster_name                                   = local.identifier
   cluster_version                                = "1.21"
   enable_irsa                                    = true
-  cluster_endpoint_public_access                 = !var.protect_resources
+  # Need public access even when deploying from AWS due to the occasional inability to access private endpoints.
+  cluster_endpoint_public_access                 = true
```
djmetzle, line 200: This requires a tfsec annotation?

ddvdozuki: No? There isn't one.

cloudprem/physical/util/dms-stop.sh
ddvdozuki: There are a bunch of statuses it could be in which is why I only check for the ones that indicate it is still running, i.e. `running` and `stopping`. Any other possible status means it is stopped and we don't need to do anything else. You can see the first version checked for "stopped" and it occasionally caused an infinite loop because the DMS task might be in one of those other statuses that's effectively "stopped" but named something else like "stopped with error".

cloudprem/physical/eks.tf
djmetzle: Hmm. I thought i remember seeing a tfsec annotation against public cluster endpoints? There's definitely a tfsec rule for that... https://aquasecurity.github.io/tfsec/v0.63.1/checks/aws/eks/no-public-cluster-access/

ddvdozuki: The reason why this doesn't get triggered is because tfsec is set to ignore imported modules. This is mostly due to 2 factors:

1. We aren't able to fix any issues tfsec might flag so it requires us to put a giant block of annotations above the module import which seems exactly as effective as not parsing the module code in the first place.
2. Tfsec is not great at parsing modules in the first place because it doesn't really handle dynamic blocks well. It will report on every possible configuration using that module and many modules allow you to deploy an insecure configuration. So even if our implementation of the module is secure, tfsec will still complain that it isn't. 

If you have a better solution or any recommendations I'd love to hear them 👍

ddvdozuki: Thought I did just run it manually on my workstation with including the downloaded modules and it didn't pickup any issues there either so still not sure what's up with that check.
Screen Shot 2022-01-29 at 2 30 39 AM

ddvdozuki: Ok yep. Just double checked and I was wrong. There is a skip notation but it's at the top of the module declaration, not at that line which is why I missed it: https://github.com/Dozuki/CloudPrem-Infra/blob/7952b9b6e56bd2c81f59553acd24430081918200/cloudprem/physical/eks.tf#L184

ddvdozuki: But you see what I mean? I took off the `#tfsec:ignore:aws-eks-encrypt-secrets` notation because I did add secret encryption here: https://github.com/Dozuki/CloudPrem-Infra/blob/7952b9b6e56bd2c81f59553acd24430081918200/cloudprem/physical/eks.tf#L203-L208 but tfsec doesn't handle modules well so it still thinks secrets aren't encrypted due to the module using a dynamic block:

Screen Shot 2022-01-29 at 2 35 36 AM

So that notation has to remain even thought we do have encrypted secrets. TFsec is great but it still has a long way to go with handling modules.

live/standard/eu-west-2/full/physical/terragrunt.hcl
```diff
@@ -0,0 +1,20 @@
...
+retryable_errors = [
+  "(?s).*error waiting for Route in Route Table.*waiting for state to become 'ready'.*"
+]
+
+# These are the variables we have to pass in to use the module specified in the terragrunt configuration above
+inputs = {
+
+}
+
+skip = get_env("SKIP_FULL", false)
```
caphene, line 20: Not a blocker, more of a question: Can this be moved up top like some of the other skip values in some of the other files?

Like in `live/standard/eu-west-2/full/logical/terragrunt.hcl`

caphene (approved):
CR 👍 

I had one question but it was more of a style question than anything else.

Good work @ddvdozuki

live/standard/eu-west-2/full/physical/terragrunt.hcl
ddvdozuki: Sure

cloudprem/physical/eks.tf
djmetzle: Gotcha

djmetzle: Just wanted to double check that.

I think the tfsec rule makes sense though. We probably dont generally want control plane access to be public. But it's also probably not a big deal for now. Not a blocker...

PR #14, opened 2022-03-15
RDS Parameter Fix

ddvdozuki:
This is a minor change to our parameter groups to add `group_concat_max_len` parameter to better match with our production cloud configuration. The lack of it was causing an issue with our customers data when migrating from production cloud as we use Persona in production and vanilla MySQL in private cloud. 

Signed-off-by: David Della Vecchia <ddv@dozuki.com>

Fixes https://github.com/iFixit/dozuki/issues/3108


scbarber (approved):
CR :barber:

caphene:
CR 🎸

PR #18, opened 2022-05-24
Lock kubectl utility to known good version.

ddvdozuki:
The `kubectl` utility had an update that made it incompatible with the current available versions of EKS which ended up breaking our bastion's ability to access customer EKS instances. This change locks the `kubectl` version to a known good one instead of just always downloading the latest version.

Additionally I've done some fixing of the tfsec and reviewdog configuration in the GitHub action workflow to improve it's functionality and reliability.

### DIFF
--- a/cloudprem/physical/static/bastion_userdata.yml
+++ b/cloudprem/physical/static/bastion_userdata.yml
@@ -28,7 +28,7 @@ write_files:
     #!/bin/bash -xe
     set -o xtrace
 
-    curl -L curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.23.6/bin/linux/amd64/kubectl \
+    curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.23.6/bin/linux/amd64/kubectl \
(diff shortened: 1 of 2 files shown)


cloudprem/physical/static/bastion_userdata.yml
```diff
@@ -28,7 +28,7 @@ write_files:
     #!/bin/bash -xe
     set -o xtrace
 
-    curl -L https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl \
+    curl -L curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.23.6/bin/linux/amd64/kubectl \
```
scbarber, line 31: `curl -L url -L0`? Is that an intentional double curl?

ddvdozuki: Yes... uh... I was just checking to make sure you were paying attention.. 😂

scbarber (approved):

PR #21, opened 2022-06-17
Updating RDS CA cert for AWS GovCloud.

ddvdozuki:
AWS went and expired the RDS CA certs without notification so our app broke for govcloud customers. This updates the cert used for the RDS instances and the app itself. We're using the bundle now so it should work moving forward even if they expire an intermediary authority.

Signed-off-by: David Della Vecchia <ddv@dozuki.com>


caphene (approved):
LGTM

scbarber (approved):

PR #22, opened 2022-06-17
RDS Parameter Group Duplication Fix

ddvdozuki:
This is to fix a race condition with our RDS parameter groups. We have certain DB parameters that need to be modified from the default so we create a parameter group with these settings however the RDS terraform module creates it's own parameter group by default as well with all the default settings and sometimes it decides to use that group instead of the group we created. We now disable this functionality and just tell it to use the existing parameter group.


caphene (approved):
LGTM 👍

scbarber (approved):

PR #24, opened 2022-07-01
BI Database VPN Access & VPN Module

ddvdozuki:
# Background
Customers will likely need access to their BI database from outside the AWS network. Our existing configuration relies on their account having an existing transit gateway managed outside the terraform. 

# Solution
We will now conditionally provision a VPN endpoint to the VPC which will allow access to authenticated users. The VPN configuration is such that only the BI database will be accessible from within the VPN and the user may configure which CIDRs they want to allow to connect to the VPN. Our VPN uses mutual authentication which relies on 2048 bit RSA self-signed TLS certificates. We generate and upload the certificates embedded in an OpenVPN configuration file to a secure S3 bucket for distribution. The bucket name and BI Database secret containing the connection information is output after the successful application of the physical module.

# QA
To test this change, download the `bi-vpn` branch, navigate to a `bi` testing environment folder in your account and region of choice, be sure the `bi_vpn_access` variable is set to `true` and the `bi_access_cidrs` variable is set to a cidr you can connect from and then apply the stack as usual. 

Once application is complete you can query the outputs for the physical module, grab the `bi_database_credential_secret` and `bi_vpn_configuration_bucket` values. The secret contains the mysql endpoint and credentials for the bi database and the configuration bucket contains the OpenVPN configuration file. Download that file to your workstation and load it into your VPN program of choice and connect. Once successfully connected you should be able to make a MySQL connection to the BI database using the credentials in the secret.

# Notable Changes
Along with the main VPN additions there are some other minor fixes as well including the following:

1. Complete S3 bucket names
   * We no longer use generic bucket prefixes, we will use the identifier and region to name the buckets something that can be easily found in the S3 dashboard
2. DMS stop script race condition mitigation
   * There seems to be some lag in the API between when a DMS task is stopped vs when it's actually reported as stopped in subsequent API calls so I added a 5 minute sleep to the end of the script to give the API time to update. This seems to have fixed this issue in my tests.
3. Bastion broken Curl & Helm Install
   * kubectl install has been broken so I fixed that. I also added helm install since that is a helpful utility to have available for debugging.
4. Tighter security for databases
   * To facilitate secure VPN usage I tightened up the security groups for the databases. We now only allow access to what's necessary instead of just allowing the entire VPC.

### DIFF
--- a/cloudprem/physical/modules/vpn/acm_ca.tf
+++ b/cloudprem/physical/modules/vpn/acm_ca.tf
@@ -1,4 +1,3 @@
-# AWS ACM certificate CA
 resource "tls_private_key" "ca" {
   algorithm = "RSA"
 }
@@ -18,7 +17,6 @@ resource "tls_self_signed_cert" "ca" {
   ]
 }
 
-# AWS ACM certificate
 resource "aws_acm_certificate" "ca" {
   private_key      = tls_private_key.ca.private_key_pem
   certificate_body = tls_self_signed_cert.ca.cert_pem
@@ -32,14 +30,10 @@ resource "aws_ssm_parameter" "vpn_ca_key" {
   description = "VPN CA key"
   type        = "SecureString"
   value       = tls_private_key.ca.private_key_pem
-
-  #  tags = local.tags
 }
 resource "aws_ssm_parameter" "vpn_ca_cert" {
   name        = "${local.ssm_prefix}/acm/vpn/ca_cert"
   description = "VPN CA cert"
   type        = "SecureString"
   value       = tls_self_signed_cert.ca.cert_pem
-
-  #  tags = local.tags
 }
--- a/cloudprem/physical/modules/vpn/acm_ca.tf
+++ b/cloudprem/physical/modules/vpn/acm_ca.tf
@@ -1,0 +1,45 @@
+# AWS ACM certificate CA
+resource "tls_private_key" "ca" {
+  algorithm = "RSA"
+}
+resource "tls_self_signed_cert" "ca" {
+  key_algorithm   = "RSA"
+  private_key_pem = tls_private_key.ca.private_key_pem
+
+  subject {
+    common_name  = "${local.identifier}.${data.aws_region.current.name}.vpn.ca"
+    organization = local.identifier
+  }
+  validity_period_hours = 87600
+  is_ca_certificate     = true
+  allowed_uses = [
+    "cert_signing",
+    "crl_signing",
+  ]
+}
+
+# AWS ACM certificate
+resource "aws_acm_certificate" "ca" {
+  private_key      = tls_private_key.ca.private_key_pem
+  certificate_body = tls_self_signed_cert.ca.cert_pem
+
+  tags = local.tags
+}
+
+# AWS SSM records
+resource "aws_ssm_parameter" "vpn_ca_key" {
+  name        = "${local.ssm_prefix}/acm/vpn/ca_key"
+  description = "VPN CA key"
+  type        = "SecureString"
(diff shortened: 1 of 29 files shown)


cloudprem/physical/modules/vpn/config.tf
```diff
@@ -0,0 +1,117 @@
...
+
+  #  tags = merge(
+  #    local.tags,
+  #    {
+  #      Name         = "VPN ${var.vpn-client-list[count.index]} client key imported in AWS ACM"
+  #      Tier         = "Private"
+  #      CostType     = "AlwaysCreated"
+  #      BackupPolicy = "n/a"
+  #    }
+  #  )
```
scbarber, line 98: Similar question from earlier: tag or no tag?

cloudprem/physical/modules/vpn/endpoint.tf
```diff
@@ -0,0 +1,40 @@
+# AWS vpn client endpoint
+resource "aws_ec2_client_vpn_endpoint" "vpn-client" {
+  description            = "${local.identifier}-vpn-client"
+  server_certificate_arn = aws_acm_certificate.server.arn
+  #  vpc_id                 = var.vpc_id
+  #  security_group_ids     = [aws_security_group.vpn.id]
+  client_cidr_block = var.client_cidr_block
+  #  session_timeout_hours  = var.session_timeout_hours
```
scbarber, line 8: Can we clean up these comments? Months from now I won't know if these are left-over from testing or placeholders or... :)

scbarber (approved):
Overall it looks good. There are a few sections of commented out code that might end up adding confusion to future maintenance efforts, and a reference to UDP port 443 which struck me as non-obvious which might need to be commented/documented (unless it is obvious and I just need more training).

cloudprem/physical/modules/vpn/acm_ca.tf
```diff
@@ -0,0 +1,45 @@
...
+}
+
+# AWS SSM records
+resource "aws_ssm_parameter" "vpn_ca_key" {
+  name        = "${local.ssm_prefix}/acm/vpn/ca_key"
+  description = "VPN CA key"
+  type        = "SecureString"
+  value       = tls_private_key.ca.private_key_pem
+
+  #  tags = local.tags
```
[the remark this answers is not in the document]
ddvdozuki: The SSM parameters get tagged implicitly automatically so tagging them here causes a conflict that prevents creation. I commented them out temporarily while I investigated and it should not have made it into the PR. I submitted a new commit that cleaned them all up

cloudprem/physical/modules/vpn/endpoint.tf
ddvdozuki: Yep. They should've been cleaned up, good catch.

cloudprem/physical/modules/vpn/sg.tf
```diff
@@ -0,0 +1,28 @@
...
+resource "aws_security_group" "vpn" {
+  name        = "${local.identifier}-vpn-security-group"
+  description = "Allows access to connect to the VPN and unlimited egress."
+
+  vpc_id = var.vpc_id
+  ingress {
+    description = "Allow VPN connection from specified CIDRs"
+    from_port   = 443
+    to_port     = 443
+    protocol    = "udp"
```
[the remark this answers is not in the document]
ddvdozuki: AWS VPN Endpoints default to using port 443. You can configure them to use 1194 instead but I usually opt to stick with the more ambiguous port numbers for security. 1194 is only openvpn but 443 could be anything. If you think it wiser to use 1194 we can definitely change that.

scbarber (approved):

PR #26, opened 2022-07-05
Terratest Suite Implementation

ddvdozuki:
It's been a long time coming but we've needed an automated test suite for our infra and at the implementation of the VPN system for BI our infra permutations became too great to continue testing each one manually. 

This test suite is written in Go using terratest by Gruntwork.io. If you want to run the tests locally there are a few steps to get started:

1. Go into the `live/tests/src` directory and run the following `go mod edit -replace dozuki.com/tests/common=../common`. As our package is not yet "deployed" in the golang sense we need that command to tell go that our test common package is actually local and do not try to download it
2. In that same directory run `go mod tidy` this will download the requisite go modules.
3. Go to `live/tests/common` and run `go mod tidy` as well. 

To actually run the test you can use `go test -v -count 1 -timeout 300m ./...` from the `live/tests/src` directory. Be aware however that this will generate a ton of interlaced test output that is basically useless to use as a debug tool so I pass it through a deinterlacer script: `go test -v -count 1 -timeout 300m ./... |terratest_log_parser --outputdir test_output`. This will give you little to no console output but will instead deinterlace on the fly as the parser picks up the output and write individual log files for each test condition in the `test_output` folder as well as a summery file of results. You'll need to install the terratest_log_parser as described [here](https://terratest.gruntwork.io/docs/testing-best-practices/debugging-interleaved-test-output/#installing-the-utility-binaries) .

And for the flags: 
* `-v` is for verbose, without that you will only get the test condition pass/fail output
* `-count 1` is to force go to run each test, go by default wants to cache tests if the go code hasn't changed but our test files are unlikely to change during development we must short circuit this behavior.
* `-timeout 300m` should be self-explanatory but as to why 300m is the time I chose, my test suite runs have been coming in at around 90-100 minutes so putting the timeout at 300 should be more than enough time. Also the default timeout is 10m  so this flag is necessary.
* `./...` this says to run all test conditions (files that end in `_test.go`) in all directories below this one.

Obviously you'll need golang installed, this has been tested with go1.18 on darwin/arm64 but should work on any system that supports that version of go.

Note: There was a change needed to our parent `terragrunt.hcl` file to allow for an environment based state key so you'll see a `TG_STATE_PREFIX` variable. This is meant **_only_** for test usage and should not be used or relied on for actual deployments. Using gitops and per-folder deployments keeps everything straight and auditable but for automated testing purposes this is infeasible.


caphene:
Just to clarify: These tests actually deploy the infrastructure?

caphene (approved):
LGTM

Nice work, @ddvdozuki

scbarber (approved):

PR #35, opened 2022-07-29
Use SSM documents for bastion configuration

ddvdozuki:
# Background
We originally used userdata and cloudinit to configure our bastion server. This works for 99% of our customers but leaves an important gap in our flexibility. Should the database hostname or ekscluster name change at some point (like during a stack migration) the bastion's configuration will not be updated due to the fact that userdata is only ever applied once at instance start. Additionally we bake-in the hostname and cluster name into the userdata so changing those requires changing the userdata itself which is somewhat useless due to the aforementioned issue with the single application of said userdata.

# Solution
SSM Documents

### Document
Since we use Systems Manager to connect to our bastion server it made sense that we would use some of their other powerful features to close this gap. Notably their "document" functionality. You can think of these "documents" as encapsulated automations, with their own prerequisites, inputs, and actions. Notated in YAML or JSON, documents are created at any time and have no dependency on anything else, they are simply waiting for an association to one or more instances. (see [cloudprem/physical/static/bastion_kubernetes_config.yaml](https://github.com/Dozuki/CloudPrem-Infra/pull/35/files#diff-f6b4ebfc59ac97053e455f377a3e1ceb43a1e6647dfe34d529cda48855364757) for an example)

### Association
The other half of this functionality is the association. Documents must be associated with one or more instances to actually do anything. It's at association time that you input the parameters to customize the functionality to your needs. This is also where you direct the document to its targets. You can use raw instance ids or in our case, tags. We tag our bastion the same as every other resource with Environment and Identifier, as well as an additional tag Role which in this case equals "Bastion". This effectively links up the environmental information with the static automation. This allows us to be flexible and update the association any time the underlying information changes, and with the way SSM documents work, a new association means a new invocation so it will automatically re-configure the bastion. (see [cloudprem/physical/bastion.tf](https://github.com/Dozuki/CloudPrem-Infra/pull/35/files#diff-b36569de93a9bfc5e4ae676b2418ddfb46a0c36367c0582c0cf7cfa545f87461) for an example of an association)

If you'd like further reading on SSM Documents, here's an actual useful bit from AWS' own documentation: [https://docs.aws.amazon.com/systems-manager/latest/userguide/document-schemas-features.html](https://docs.aws.amazon.com/systems-manager/latest/userguide/document-schemas-features.html)

# QA
To test this change, download the `bastion-improvements` branch, navigate to the `live` directory and deploy a stack on one or more accounts/environments. Once deployed login to system manager in the console GUI, start a new terminal session with the bastion, and verify that `kubectl`, `helm`, and `mysql` are all present and configured properly. Additionally you can run any one or all of the test suite as they all now check for the existence and in some cases the configuration of these utilities.

# Notable Changes
There are some seamless upgrades made to the infra and test suite in addition to the bastion improvements:

1. Upgrade the DMS engine version from 3.4.5 to 3.4.6. I received an email from AWS warning that 3.4.5 would no longer be supported soon for new instances so better to update now.
2. No longer conditionally create primary rds parameter group. Conditionally creating this parameter group made it impossible to upgrade from a base installation to one with a BI server because terraform would attempt to delete this parameter group while it was still attached to the primary RDS server which resulted in a failure.
3. Added some test-related utility scripts. `clean.sh` and `run_tests.sh`. The run tests script runs the entire test suite and puts all test output into a timestamped folder. It also runs the clean script after test completion. The clean script depends on `awsweeper` being installed to clean the test environment of all deployed resources. **Be sure to update the AWS profile if you do not use the "default" profile for the test account as I do.** 
4. A fairly sizable refactor of the test suite assertions. In addition to adding bastion tests, we now group all common assertions into one function so they can be executed with one line in the test suites themselves. I've also made changes to ensure govcloud support for the AWS utilities used during testing (the built-in terratest aws utilities do not support govcloud 🤦 )
5. We now have an exclusion that runs on the govcloud tests to prevent attempting to deploy instances in govcloud that are not supported in all us-gov-west-1 AZ's. This is something that should eventually handled in the terraform but for now we handle it here.


scbarber (approved):
LGTM

caphene (approved):
LGTM

Nice work @ddvdozuki 👍

PR #36, opened 2022-08-18
Fix upgrading from older versions (<=2.5.4)

ddvdozuki:
We had some changes to the terraform that was breaking backwards compatibility and causing terraform to try to delete and re-create resources erroneously. This fixes that by using `moved` blocks to tell terraform which resources we renamed as well as lifecycle rules to ignore changes where appropriate.


scbarber (approved):

caphene (approved):
LGTM 👍

scbarber (approved):

PR #37, opened 2022-08-19
Webhooks DB Update Fix

ddvdozuki:
Creating the frontegg database AND updating the sites db to enable webhooks in the same k8 job makes the entire process very fragile and susceptible to order of operations issues. This splits those and does them when more appropriate. The frontegg db is created as soon as we have a database connection whereas the sites db update waits for the app to be ready (and therefore the sites db actually existing) before trying to update it.


caphene (approved):
LGTM 👍

PR #42, opened 2022-09-27
Add SSL to Grafana frontend

ddvdozuki:
My MVP of Grafana did not include SSL access for the frontend, this PR adds that by using the self-signed cert functionality of the VPN module. I refactored said functionality into its own module, now stored in the `common` root directory, and is used by _both_ the VPN module _and_ the Grafana implementation. Additionally I refactored the way I set Grafana config directives: I no longer use the `grafana.ini` option in the helm chart, opting instead for environment variables set via a k8 secret.

### DIFF
--- a/cloudprem/common/acm/ca.tf
+++ b/cloudprem/common/acm/ca.tf
@@ -1,3 +1,4 @@
+
 resource "tls_private_key" "ca" {
   algorithm = "RSA"
 }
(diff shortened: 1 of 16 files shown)


scbarber (approved):
Nicely done!

ddvdozuki:
Re-verified tflint changes and upgradeability from previous release

scbarber (approved):

PR #44, opened 2022-10-05
Restricted Access SG Fix

ddvdozuki:
There was an edge-case issue in the infra where if a customer set a restricted access CIDR for the app, dashboard, or Grafana, it would also disable the ability for the NLB to do health checks on the instances thereby setting the instances as "unhealthy" and shutting down access to the endpoint entirely after a certain amount of time. This slipped through our testing due to the 2-5 minute lag between instance creation and the health checks failing enough to disable them.

We now detect the CIDR set in the config and if it's restricted we append the local VPC CIDR to the list ensuring the health checks will always have access.

I've also added some minor refactoring in this PR as well. Some housekeeping done on the Grafana dashboard config (we were still trying to load a deleted dashboard which was filling the logs with errors) and some updates to our test suite to keep it in parity with our tf.


live/clean.sh
```diff
@@ -11,36 +11,41 @@ cleanPrefixes() {
...
+#                 pushd $region
+#                 for env in ${envs[@]}; do
+#                     echo "Clearing $env"
+#                     pushd $env/physical
+#                     TG_STATE_PREFIX="test/$p/" terragrunt state rm "module.eks_cluster.kubernetes_config_map.aws_auth[0]"
+#                     popd
+#                 done
+#                 popd
+#              fi
+#         done
```
scbarber, line 26: Temporarily commented out?

scbarber (approved):

live/clean.sh
ddvdozuki: More like it's there as an option if you want to enable it. Sometimes it's necessary but it adds quite a bit of time to the clean script so I usually have it commented out unless I need it.

caphene (approved):
Nice.
Good work, @ddvdozuki!

PR #49, opened 2022-10-14
Fix missing min size for worker AutoScalingGroup

ddvdozuki:
Somewhere during the EKS module refactor we lost the minsize parameter. I think I felt overconfident in the scaling system to choose the minsize on its own which is why I left it at the default of 1. This fixes that to allow us to tweak the minsize and guarantee capacity if needed.

Additionally I updated the cluster-autoscaler helm chart to the latest version.


caphene (approved):
Approved! 👍

scbarber (approved):

PR #55, opened 2022-12-13
Add the saml.json config

scbarber:
This is needed to configure the app to use the SSL cert to sign SAML requests.


ddvdozuki:
Looks good as long as those files are indeed present in the docker image 👍

ddvdozuki (approved):

PR #57, opened 2023-01-20
Refactor logical module to use KOTS instead of replicated scheduler helm chart

ddvdozuki:
These are all the necessary terraform changes to deploy our app using KOTS instead of the replicated kubernetes scheduler. This is a first effort and far from ideal. We will need to iterate on this to iron out the kinks and return to full feature parity (at least as much as is supported by replicated)

This should be considered an alpha release.

## Limitations:

* Admin dashboard cannot be accessed via NLB (additional k8s service required and I was not able to get it to work, not worth holding this up)
* Replicated sequence targeting no longer supported, at least not in the same way as it was before so it has been temporarily removed
* Self-signed SSL certs are the _only_ type currently supported. To validate SSL is working I needed self-signed certs but with the way KOTS no longer auto-generates a cert, I had to generate one from within terraform leaving no room for a `www-tls` secret definition in the replicated manifest. This means a proper cert **cannot** be installed (this is top of the list of things to fix)
* I ran into some issues with the new replicated license format. They are now in YAML format and are more akin to k8s CRDs. They include much more information than the old licenses. Once I added all the existing license fields (govern, modern tls, users, webhooks, etc) the license size grew too large for AWS parameter store (parameters are limited to 8kb and the replicated licenses are now 9 KILO BYTES of text). I had to cut some stuff out and remove those license parameters for it to work. Right now we only have webhooks related parameters. We will need to find a new solution for parameter storage, I plan on asking replicated directly if they have some sort or programmatic way to retrieve licenses and perhaps we can just use customer keys instead?
* Due to time constraints I have not updated the test suite to reflect these new changes.

## Deploying

Currently this only deploys via workstations. I am still in the final stages of wrapping up the codebuild changes necessary to deploy KOTS, namely getting kubectl installed in the codebuild environment and configuring said environment to connect to the eks cluster. This should be done in the next day or two but I felt it important to get this PR out ASAP. If you intend to deploy environments from your workstation be sure to have kubectl installed as well as kots. The terraform _will_ attempt to install kots if it's not there but if you are on a mac it will fail due to not having sudo privileges.

## Notes

Additionally I refactored the directory name for the terraform from `cloudprem` to `terraform`. This has been something that has annoyed me from day 1 and needed to be fixed sooner rather than later.


live/standard/us-east-1/bi/env.hcl
```diff
@@ -6,10 +6,11 @@ locals {
   enable_bi = true
   rds_multi_az = false
   highly_available_nat_gateway = false
-  dozuki_license_parameter_name = "/dozuki/workstation/beta/license"
+  dozuki_license_parameter_name = "/dozuki/workstation/kots"
   protect_resources = false
   bi_public_access = false
   bi_access_cidrs = ["0.0.0.0/0"]
   grafana_access_cidrs = ["0.0.0.0/0"]
+#  grafana_ssl_cn = "ccswb.dozuki.com"
```
scbarber, line 14: Left over debugging comment?

live/standard/us-east-1/min/env.hcl
```diff
@@ -2,10 +2,11 @@
 # feed forward to the child modules.
 locals {
   environment = "min"
+  #identifier = "a234567890"
```
scbarber, line 5: Left over debugging comment?

scbarber (approved):

caphene (approved):
Good work @ddvdozuki 👍

PR #60, opened 2023-02-10
Restore feature parity and improve reliability of KOTS deploys

ddvdozuki:
# Overview
This PR will restore most of the features lost with the KOTS upgrade:

1. Working Grafana, Frontegg & Webhooks
3. Dynamically setting the Hostname to the NLB
4. Replicated dashboard access
5. License install during bootstrap

## Grafana & Frontegg

Because of the way KOTS deploys we can no longer make use of cross-namespace communication in kubernetes (we never should've had to but that's another argument). We now put all relevant resources into the created `dozuki` namespace. 

## NLB Hostname

The old replicated scheduler allowed for a bootstrap config file. If it existed in the right place the control plane would pick it up automatically and apply the settings. Within the file was the ability to set the dashboard hostname which would be inherited by the app. Thankfully kots does things a little more predictably and a little more inline with K8s standards. We now have to create a yaml manifest with config values declared and pass that to the kots installer. We can set any config values we want this way but right now we are only using it for the hostname.

## Dashboard Access

By default KOTS does not allow direct access to the admin dashboard. You must execute a `kubectl` command to open a local proxy. This obviously will not work for us due to some customers needing access so I had to create my own K8s service to route traffic to the admin dashboard manually. Unfortunately since the KOTS dashboard does not support TLS the connection is currently insecure. In an upcoming PR we will implement an SSL termination ingress controller to handle this.

## License Install

This is the big one. Our old replicated licenses were fairly small. Usually just 1 string of around 200 characters of base64 encoded json. And within the json blob was a key fingerprint and a customer ID. This was fed to the replicated control plane and then it called out to replicated servers directly to query for license details. This had the added benefit of protecting our secure license fields that we don't want exposed to customers. Unfortunately KOTS licenses are not only fully k8s CRDs, but they include, in plaintext, all of our secure fields making it impossible to share the license data directly with the customers. 

So we have 2 problems:

1. Physical license size. Our KOTS licenses are now over 9kb in size and the maximum amount of text that can be entered into an AWS parameter is 4kb. 
2. Plaintext license fields. Our webhook connection information is exposed in the license file, not even base64 encoded. Theoretically anyone with these credentials could add or remove webhooks to all our clients.

The solution to both problems is the same: we don't give the license to the customer. Thankfully replicated has an API endpoint we can hit with a curl request to pull down the license manifest with only a customer ID needed as an authorization field. This isn't the most secure thing in the world but for now I feel It's good enough until we move to other deployment strategies in the future. The customer ID is a UUID so it should be fairly safe as long as it remains hidden.

# Leftovers
There are 2 main features that are left to be implemented in a future PR:

1. SSL termination for the admin dashboard
2. Custom SSL cert support for the app itself.

### DIFF
--- a/cloudprem/logical/charts/kubed/OWNERS
+++ b/cloudprem/logical/charts/kubed/OWNERS
@@ -1,5 +1,0 @@
-approvers:
-- tamalsaha
-reviewers:
-- tamalsaha
-
--- a/live/gov/us-gov-west-1/bi/env.hcl
+++ b/live/gov/us-gov-west-1/bi/env.hcl
@@ -6,7 +6,7 @@ locals {
   enable_bi = true
   rds_multi_az = false
   highly_available_nat_gateway = false
-  dozuki_license_parameter_name = "/dozuki/workstation/kots"
+  dozuki_customer_id_parameter_name = "/dozuki/workstation/kots/default/customer_id"
   protect_resources = false
   bi_public_access = false
   bi_access_cidrs = ["0.0.0.0/0"]
--- a/live/gov/us-gov-west-1/full/env.hcl
+++ b/live/gov/us-gov-west-1/full/env.hcl
@@ -6,6 +6,10 @@ locals {
   enable_bi = true
   rds_multi_az = true
   highly_available_nat_gateway = false
-  dozuki_license_parameter_name = "/dozuki/workstation/webhooks/kots"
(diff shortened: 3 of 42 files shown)


scbarber (approved):
LGTM! Nice work.

scbarber (approved):

PR #61, opened 2023-02-19
Module & Live Environment Refactor with SSL Changes

ddvdozuki:
# Overview
The last thing left to ensure feature parity for KOTS was custom SSL certificate support and SSL support for the replicated dashboard. This PR includes necessary changes for both of those fixes though most of the changes were made to the replicated release itself. In addition some much needed refactoring was done to the live environment and the main root modules themselves. Here is a list of what has changed:

1. No more copied live environments for each region and partition. Instead we have a skeleton directory with the DRYest version of the terragrunt config and a simple shell script to generate the previous live environment from said skeleton directory. This allows us to ensure changes are uniform across all regions and partitions and cuts down on copied code.
2. We have certain resources that need to exist on any new account we create. We've been creating these manually but it's time we codify it properly so i've created a `bootstrap` folder that includes the terraform necessary to setup any SSM parameters needed for workstation deploys as well as the dreaded `dms_vpc` role.
3. The introduction of custom grafana SSL certs last year inadvertently broke our physical/logical module system by tasking the logical module with a managing a physical resource in the form of the SSM and ACM resources needed for the certificate and key. Adding SSL generation for the NLB and replicated dashboard exasperated this and made it impossible for a stack to be cleanly deleted when created via CloudFormation so resources were moved to maintain this functionality.

## Live Environment
Our previous live environment consisted of 2 partition folders (`gov`, and `standard`) as well as one folder for each region in said partition and yet more folders inside each region folder for all the supported configurations (`bi`, `webhooks`, `full`, and `min`) giving us a tremendous amount of copied folders and files. Any time one would need to make a change to the terragrunt or environmental config one would have to ensure the change was copied successfully to each of a dozen or more files. This list of files would of course grow as more regions were added. 

Instead I created a `live/.skel` directory that houses 1 copy of the configurations mentioned above with the relevant environment file for each. That way if you need to change something for the `bi/env.hcl` file it will change for every region and partition. I've also created a `regions.txt` file in the `live/.skel` directory that informs the shell script how many region folders to create. This file can be added to or removed as our region support changes. 

To generate the live environment, navigate to the `live` folder and run `./generate_live_env.sh`. This can be run any time changes to the skel are made to re-generate the live environment. I've also added the `live/gov` and `live/standard` folders to the gitignore to prevent them from being committed again in the future.

## Bootstrapping
To deploy via a workstation and not the CloudFormation we need to ensure resources that are usually created by CF are available to the terraform. Most notably the SSM parameter for the Customer ID. These are parameters that need to exist in every region that we support on any account you wish to deploy via the workstation. 

I added a small bit of terraform in the `terraform/bootstrap` folder that allows you to deploy these parameters easily on a new account via terragrunt. To run this code you would navigate to the live environment and partition folder (once generated) and execute `SKIP_INFRA=true SKIP_SETUP=false terragrunt run-all apply -var 'customer_id_parameters={default: "abc12345", webhooks: "def6789"}'`. This will tell terragrunt to skip deploying the actual infra and only deploy the bootstrapping code for each region folder in the live environment. 

In addition to the SSM parameter creation for workstations the bootstrap code will also deploy the `dms_vpc` role for new accounts. You can execute that in a similar way to the parameters: `SKIP_INFRA=true SKIP_SETUP=false terragrunt run-all apply -var 'dms_setup=true'`. Remember this only needs to be executed once **per account** and only for customers that need BI.

## SSL & Module Refactoring
We haven't been bitten by the left-over grafana SSL issue yet due to it being an optional and as-yet unneeded code in the terraform. However now that we are generating an SSL cert for the NLB/Replicated dashboard on _every_ deploy, this is now the well-trodden path and needs to be fixed.

The issue is in the way we destroy stacks in CodePipeline. For simplicity and time our CodePipeline is configured to only destroy the **physical** module. Because the logical module only deploys resources _onto_ the physical module deleting the physical resources deletes the logical ones implicitly. It's been a great system that has operated nearly flawlessly since its inception (thanks metz for that idea). Unfortunately in my haste to implement custom SSL cert generation for grafana last year I put the terraform responsible for managing that resource into the logical module. This code creates numerous SSM parameters and ACM certificates which for our purposes are considered very physical resources. So if this stack with custom SSL generation for grafana was to be deleted by CodePipeline it would leave behind a bunch of resources that would prevent a new stack with the same environment name/identifier from being created without those resources being manually deleted. All in all, very dirty; and not at all what what I envisioned for repeatability and testability through iterative continuous deployment.

Enter KOTS. With it completely removing self-signed cert generation from it's control plane we must now do this ourselves via terraform. My initial instinct was to just put it right next to the grafana cert generation code but I mysteriously found myself unable to rapidly re-create stacks in the same environment. After a short investigation I realized my mistake. And since this is not a rarely used path like the grafana ssl one but instead the **default** path... well you can see why it needed to be refactored. 

It required piping parameter names through the physical module into the logical module which isn't a big deal but the plumbing needed to be done and then a new round of testing to ensure things still worked as intended. All in all I believe the work was a good investment to retain our single module delete functionality.

## Replicated Release Changes
As said above much of the support for custom SSL and dashboard SSL was added to the replicated release and not the terraform. Quite a lot was done there including adding an entire ingress support tree of resources and correctly piping the SSL cert from the dashboard into the app container. See the `ingress.yaml` and `secrets.yaml` files in the release for details.


scbarber (approved):
LGTM. Thanks DDV!

PR #63, opened 2023-03-01
Fix missing env.hcl issue with bootstrap config in live environment

ddvdozuki:
We came across this weird issue where the terragrunt function `find_in_parent_folders()` really means parent, like EVERY parent, all the way to `/`. Due to this it's possible for a user (like me) to have a rogue `env.hcl` file in your home directory and it still get included in the parsing of the module if no other env.hcl file is found first. Of course if you DON'T have a spare `env.hcl` file in your home directory (i mean why don't you have one?) then the run-all command will fail.

This PR fixes that issue by refactoring the directory structure a little bit with the workstation setup module and adding an env.hcl file in the appropriate location. It also has the added benefit of providing semantic instruction on what variables to use when running the bootstrap.

Note: you will need to re-generate your live env locally once you pull down this change by running the `generate_live_env.sh` script in the live folder


caphene (approved):
LGTM 👍

scbarber (approved):

segrettodozuki (approved):
not too much going on here - looks good to me.

PR #66, opened 2023-03-06
Feature/ops 308

segrettodozuki:
This code change will 
  1.  upgrade the bastion AWS CLI from v1 to v2
  2. upgrade the bastion kubectl client with the lastest patches of the same minor version.
  3. install k9s on the bastion

K9s requires the AWS CLI v2 so that the 'aws eks' command we use to generate the kubeconfig, generates a kubeconfig that k9s can use to connect to EKS.  This is the best fix to the issue describred here:

https://github.com/derailed/k9s/issues/1619

In short k9 was built with a recent k8s.io/api package and the AWS cli v1 will not produce a proper kubeconfig that supports that api package.

To update your bastion with installing K9s run the following:

terragrunt run-all apply -replace=aws_ssm_association.bastion_kubernetes_config -replace=module.bastion.aws_launch_configuration.this[0]

That will force a rebuild of the bastion and force a rerunning of the SSM scripts.

### DIFF
--- a/terraform/physical/static/bastion_kubernetes_config.yaml
+++ b/terraform/physical/static/bastion_kubernetes_config.yaml
@@ -22,6 +22,18 @@ mainSteps:
       sourceInfo:
         url: "https://dl.k8s.io/release/v1.21.0/bin/linux/amd64/kubectl"
       destinationPath: "/usr/bin/kubectl"
+  - action: aws:runShellScript
+    name: installK9s
+    inputs:
+      runCommand:
+        - mkdir k9stemp
+        - cd k9stemp
+        - curl -LO https://github.com/derailed/k9s/releases/download/v0.27.3/k9s_Linux_amd64.tar.gz
+        - tar -xzf k9s_Linux_amd64.tar.gz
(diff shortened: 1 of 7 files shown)


terraform/physical/static/bastion_kubernetes_config.yaml
```diff
@@ -15,13 +15,35 @@ parameters:
...
       destinationPath: "/usr/bin/kubectl"
+  - action: aws:runShellScript
+    name: installK9s
+    inputs:
+      runCommand:
+        - mkdir k9stemp
+        - cd k9stemp
+        - curl -LO https://github.com/derailed/k9s/releases/download/v0.27.3/k9s_Linux_amd64.tar.gz
+        - tar -xzf k9s_Linux_amd64.tar.gz
+        - sudo cp k9s /usr/local/bin/; sudo cp k9s /usr/bin/
```
scbarber, line 44: Curious: what's the value in copying the binary to 2 locations?

segrettodozuki:
closing this in group agreed preference for https://github.com/Dozuki/CloudPrem-Infra/pull/67

PR #67, opened 2023-03-06
Feature/ops 308b

segrettodozuki:
This is probably a better way of satisfying the dependency requirements of k9s for eks on the bastion than what I did in the other not merged PR https://github.com/Dozuki/CloudPrem-Infra/pull/66.

1.  upgrades OS from AL2 to AL2023
2. upgrades mariadb from 5.5 to 10.5 (so we can continue to use packages to install mariadb)
3. installs latest patches for same kubectl minor version
4. installs latest stable k9s

May need further testing to prove mariadb 10.5 does not require we update our scripts that will use it.

If we merge this, then we will update our bastions with the following command:

terragrunt run-all apply -replace=aws_ssm_association.bastion_kubernetes_config -replace=aws_ssm_document.bastion_mysql_config -replace=module.bastion.aws_launch_configuration.this[0]

### DIFF
--- a/live/clean.sh
+++ b/live/clean.sh
@@ -41,7 +41,7 @@ fi
 
 echo "Cleaning standard"
 pushd standard || exit
-#cleanPrefixes
+cleanPrefixes
 for region in *; do
  if [ -d "$region" ]; then
     echo "Cleaning $region"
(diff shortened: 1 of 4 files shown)


ddvdozuki:
For the terragrunt command you have written there, you should not need to specify the ssm documents as they have changed so terraform will replace them on their own. Triggering a re-create of the launch configuration would be helpful, normally I've done this in the past by simply deleting the bastion instance itself and letting it get recreated but this is a better solution.

segrettodozuki:
> For the terragrunt command you have written there, you should not need to specify the ssm documents as they have changed so terraform will replace them on their own. Triggering a re-create of the launch configuration would be helpful, normally I've done this in the past by simply deleting the bastion instance itself and letting it get recreated but this is a better solution.

I agree with you that you shouldn't need to replace the ssm, and that update in place should be sufficient.  I tried that a little while ago, and it didn't run the new ssm command on the new bastion instance it spun up (which has been a recurring issue for me on various update action attempts).  I've had most luck replacing both.  perhaps I am missing something and theres something else going on that was causing it to not update.  And yes, mainly the point of replacing the launchConfig is to force aws to replace the EC2 instance bastion.

ddvdozuki:
So the SSM documents are applied to any EC2 instance with the right tags but it doesn't always happen instantly. It kind of gets added to a queue and run when it has time. In my experience this usually results in a lag from 1 to 10 minutes. But it will only run the documents on an instance once. So if you update the docs it will not re-run them on the same instances hence why deleting the instance or triggering a re-create on the LC is necessary for developing ssm docs.

scbarber (approved):

ddvdozuki (approved):
LGTM Good job @segrettodozuki

PR #73, opened 2023-04-20
Full LetsEncrypt support

ddvdozuki:
# Problem
We need legitimate SSL certificates for our various endpoints: the kots dashboard, grafana, and the app itself. Additionally, we'd like to limit accessing the app with non-standard ports (this causes confusion for our customers), and reduce our reliance on Apache.

# Solution
Nginx-Ingress, Cert-Manager, and LetsEncrypt. 

Kubernetes has an accepted workflow to solve our issue by using ingress objects, cert-manager, and letsencrypt. Some obstacles lie in our way however, notably the fact that letsencrypt will *not* generate certificates for amazonaws.com (the AWS autogenerated domain for ELBs). This means we have to manage domains ourselves. Because of that this PR is mostly related to managing DNS and subdomains, the majority of the _actual_ LetsEncrypt changes are in the kots-app manifests themselves. 

A brief overview of the changes:
* Autogenerate a subdomain on dozuki.cloud for each new environment.
  * Support external domains as well (i.e. dozuki.customer.com)
* Add `cert-manager` helm chart
* Remove all generated SSL certs and reconfigure to use le certs.
* Refactor Grafana to respond to `/dashboards` instead of port 3000.
* Repurpose port 80 NLB listener for ACME http challenges and ingress http->https redirect only.

### Identifier -> Customer
Instead of adding an additional variable for the customer's subdomain, I refactored the way the `identifier` works. It's now called `customer` instead and has regex validation to ensure it's appropriate to use as a subdomain. This should cause zero changes to resources that previously used identifier as the internal implementation in the terraform is the same but it's something to look out for regardless. 

When upgrading from a stack that previously used an identifier, that identifier will be used in the autogenerated subdomain, if you don't want that to happen you can use the `subdomain_override` and specify a custom subdomain instead. This override variable should only be used for this purpose. New stacks and upgrades to stacks that did not previously have an identifier should just use the `customer` variable.

### GovCloud
GovCloud does not support public DNS zones in Route53 so any GovCloud customers that want to use our domain (dozukigov.us) will need to be treated as an "external_fqdn" and DNS entries will need to be created manually pointing to the NLB domain:
* CNAME for customer.dozukigov.us
* CNAME for *.customer.dozukigov.us

Additionally there is an edge-case where if you deploy on GovCloud and *don't* specify an external domain, it will use the NLB domain instead and therefore will *not* have an SSL certificate at all (since LE will not generate one for the amazon domain) which would render the stack unreachable in a browser. I have some changes in mind to ensure this doesn't happen but I will add them to a future PR, this one is large enough already.

### Subdomain Format
I added a little sugar to the way we generate subdomains. By default it's `customer-environment-region.dozuki.cloud` however you can modify the `subdomain_format` template and change up both the order and inclusion of environment and region. For example if you wanted `customer.dozuki.cloud` you can update the `subdomain_format` from `["%CUSTOMER%", "%ENVIRONMENT%", "%REGION%"]` to `["%CUSTOMER%"]`.  Same goes for if you wanted only customer-environment or environment-customer-region. The only requirement is the parts will always be delimited by a dash `-` (this is hard coded in the terraform), and you must always include the customer value in the template, this is enforced with a validator.

## QA
You should be able to deploy this branch on any of our organizational AWS accounts. If you want to test the deploy on govcloud, be sure to set the `external_fqdn` variable to the CNAME record you manually created . 

## CR
You can safely ignore anything in the `terraform/logical/charts` directory as those changes are just me adding the cert-manager helm chart.


terraform/logical/charts/cert-manager/Chart.template.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: v1
+name: cert-manager
```
djmetzle, line 2: Nice! :medal_sports:

scbarber (approved):
LGTM

segrettodozuki (approved):

PR #74, opened 2023-05-01
Grafana MySQL Persistance & General BI Refactor

ddvdozuki:
# Problem
When we upgrade a user it's possible it will require a replacement of the Grafana helm chart. Grafana should be treated as ephemeral, however if the helm chart is deleted so is its PVC storing the customers custom configuration requiring a lengthy manual migration process.

# Solution
We leverage Grafana's ability to use a MySQL database as it's storage backend instead of the PVC. This allows the Grafana pod and helm release in general to be treated as ephemeral and can come and go as necessary to free us up for upgrades or refactors to the infra. 

### Migration Required
We will need to manually migrate all BI customers to MySQL from SQLite when we run this upgrade. There is already a github project created by the Grafana devs for this purpose as this is a common use-case. But it must be scheduled.

### Still creating a PVC
To maintain backwards compatibility I have not removed the helm configuration to create the PVC for Grafana as this would trigger a delete of the PVC before we had a chance to migrate them to MySQL. Once all customers have been moved to MySQL we can safely remove the PVC config.

### Security Changes
I've tweaked a lot of the knobs of the Grafana config to make it as secure as possible by disabling many of the services it usually provides that we have no need for. You can see the changes in the `data` block of the `logical/bi.tf` file. Most notably disabling the metrics server, and enabling secure cookies and XSS protections. (I've also enabled `samesite` which restricts loading in an iframe so if we decide to go back to that within the app then this setting will need to be disabled)

### Misc Changes

* Grafana username is now configurable. It will use the `var.customer` value as the username or `dozuki` if none is provided.
* Refactored all the BI resources in the `physical/rds.tf` file into it's own `physical/bi.tf` file as that rds file was getting mighty large. 
* Using a k8 job to create the Grafana db along with the necessary configmap for the script and secret for the mysql credentials
* Using a k8 job to wait for the ingress controller to be ready before we deploy Grafana since the helm chart creates its own ingress object and attempting to add an ingress object before the controller is ready triggers a validation error and prevents creation. In the future we may want to look into consolidating our ingress creation in one place to prevent this race condition.
* Disabled use of `engine_version` for the DMS replication instance. For some reason terraform allows you to specify an engine version while also specifying `allow_minor_upgrade` but then when AWS upgrades the instance automatically terraform tries to downgrade to the specified version in the file.... so anyway we are not specifying a version number anymore and will default to the latest.
* Proper IAM roles for DMS CloudWatch are now created so we can get logs on the DMS job to debug issues.

## QA
You should be able to deploy this branch on any of our organizational AWS accounts. If you want to test the deploy on govcloud, be sure to set the `external_fqdn` variable to the CNAME record you manually created . 

As usual I have validated that we can safely upgrade from what's currently in master (the LE stuff) to this branch.


terraform/logical/bi.tf
```diff
@@ -41,33 +41,186 @@ resource "kubernetes_secret" "grafana_config" {
...
+    GF_USERS_DEFAULT_THEME                  = "light"
+    GF_DATABASE_TYPE                        = "mysql"
+    GF_DATABASE_HOST                        = local.db_master_host
+    GF_DATABASE_USER                        = local.db_master_username
+    GF_DATABASE_PASSWORD                    = local.db_master_password
+    GF_ANALYTICS_REPORTING_ENABLED          = false
+    GF_ANALYTICS_CHECK_FOR_UPDATES          = false
+    GF_METRICS_ENABLED                      = false
+    GF_SECURITY_COOKIE_SECURE               = true
+    GF_SECURITY_DATA_SOURCE_PROXY_WHITELIST = "1.1.1.1:1" #Disable all data source proxying
```
scbarber, line 55: Point of curiosity: why this particular string value? Makes me think of Cloudflare's DNS server address. :)

ddvdozuki: It is lol. I just needed an IP address to put in that field. I could've used 192.168.1.1 i guess. Just something that cannot ever resolve to an actual user making a request. There's no way to disable the proxy entirely, all you can do is set a bogus whitelist 🤷

terraform/logical/bi.tf
```diff
@@ -41,33 +41,186 @@ resource "kubernetes_secret" "grafana_config" {
...
+    name      = "grafana-db-create"
+    namespace = kubernetes_namespace.kots_app.metadata[0].name
+  }
+  spec {
+    template {
+      metadata {}
+      spec {
+        container {
+          name  = "grafana-db-create"
+          image = "imega/mysql-client"
```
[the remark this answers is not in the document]
ddvdozuki: Ok. We've been using that imega image for the entirety of private cloud. I just added it here as well but it's been powering the webhooks db creation for as long as it's existed. I can add a tag to this to ensure it doesn't change.

terraform/logical/bi.tf
```diff
@@ -41,33 +41,186 @@ resource "kubernetes_secret" "grafana_config" {
...
+              secret_key_ref {
+                name = kubernetes_secret.grafana_mysql_credentials[0].metadata[0].name
+                key  = "password"
+              }
+            }
+          }
+          command = [
+            "sh",
+            "-c",
+            "mysql --host=$MYSQL_HOST --user=$MYSQL_USER --password=$MYSQL_PASSWORD --execute=\"$(cat /scripts/grafana-db.sql)\""
```
scbarber, line 134: The current state of `grafana-db.sql` checks out for use in an `--execute` statement. There could be a future where quoting could create a problem then we could maybe use ` < /scripts/grafana-db.sql` instead? No need to edit it here, just observing a potential land mine in the far future.

terraform/logical/bi.tf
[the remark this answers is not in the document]
ddvdozuki: I added the `10.6.4` tag which was the latest release from 2 years ago.

terraform/logical/bi.tf
ddvdozuki: Good call. That line was generated by ChatGPT and I never really liked the cat script to execute but never went back and changed it.

scbarber (approved):
LGTM

scbarber (approved):

segrettodozuki (approved):
seems to all make sense.

PR #76, opened 2023-05-04
S3 Bucket Fix & Minor Refactors

ddvdozuki:
This is built on top of the live isolated & monitoring overhaul PR. Besides that it is a change to just a few files, notably:

1. Removal of s3 bucket ACLs for guide buckets due to AWS' new default for bucket creation enforcing bucket owner object ownership and disabling of ACLs.
2. Addition of K8 RBAC permissions for the dozuki namespace to allow for listing of podnames to support the queueworkerd beanstalkd watcher sidecar
3. Addition of `rds:AddTagsToResource` permissions for the EKS cluster due to a random "bug" that occurs during app update on some clusters requiring that permission (every time the app updates it triggers an RDS snapshot and sometimes it needs to copy tags). It's not always required, just some times. I have not figured out why but in the meantime we will add the permission to fix it.


scbarber (approved):
LGTM

caphene (approved):
Nice work! 👍

PR #77, opened 2023-05-04
Isolated Run Script for Live Environment

ddvdozuki:
This is a shell script and small Dockerfile to allow developers to run multiple iterations of the terraform on their workstations without being tied to the source material. This is helpful when testing changes and upgrades to old versions.

The previous workflow for testing changes looked like:
1. Stash working changeset
2. Check out previous tag or master
3. Trigger deploy of checked out version
4. wait.......... coffee....
5. Once completed we then check out the working branch
6. apply the stash of our working changes
7. Trigger deploy of new version on top of the old version to test upgrading.
8. wait........ coffee.....
9. profit?

All the while during those wait periods you were unable to make any changes to the terraform or live environment due to the multi root module configuration we use. It might've loaded the physical module into the cache but the logical module hadn't been touched yet so no changes can be made lest you create an issue when it's time for the logical module to run. 

Now with this script the workflow looks like this:
1. Execute `isolated_run.sh` with your parameters including the region, environment, terragrunt command, git branch (or current) and some optional flags.
2. Continue working on the code or do whatever while the docker container runs your command in the background, completely isolated with it's own copy of the terraform.

### DIFF
--- a/live/Dockerfile
+++ b/live/Dockerfile
@@ -1,5 +1,5 @@
 # Use the specified base image
-FROM devopsinfra/docker-terragrunt:aws-tf-1.3.9-tg-0.44.0
+FROM devopsinfra/docker-terragrunt:aws-latest
 
 # Set environment variables for kubectl version
 ENV KUBE_VERSION="v1.23.0"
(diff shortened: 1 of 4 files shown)


caphene (approved):
LGTM 👍

PR #78, opened 2023-05-10
Conditional RDS ReadReplica DMS Swap

ddvdozuki:
# Problem
DMS is unreliable with our current schema issues. Until those issues get fixed the DMS migration task will randomly fail. Some times it takes weeks but for other customers it's failing daily. We have not yet nailed down what exactly is causing the issue, only vague hints from DMS about specific schema incompatibilities.

# Solution
We replace DMS altogether for customers that do not need public BI or conditional replication. DMS is a highly specialized tool and not necessary for 99% of our customers. This PR will simplify the infra by using an RDS read replica instead of a DMS pipeline and let RDS manage the replication automatically.

## Caveats & Additional Changes

1. New boolean variable called `dms_enabled`.
2. I've chosen to create 2 terraform RDS resources, one for RDS RR and one for DMS. Depending on the configuration (`public_bi=true` and/or `dms_enabled=true` will enable DMS) we will deploy one or the other. This gives us flexibility to move a customer from one to the other without needing to totally tear down and redeploy as we would if we were morphing a single instance. (Read Replica is default)
3. We now handle missing `dms-vpc-role` issues with a conditional null resource provisioner. If the role exists it won't be created. This is a great way to trick terraform into a "create if not exists" solution.

## QA
You should be able to deploy this branch on any of our organizational AWS accounts. If you want to test the deploy on govcloud, be sure to set the `external_fqdn` variable to the CNAME record you manually created . Set `dms_enabled` or `public_bi` to true to test DMS otherwise a read replica will be deployed instead. 

As usual I have validated that we can safely upgrade from what's currently in master to this branch.


segrettodozuki (approved):
Looks pretty solid to me.  This should really help!

caphene (approved):
LGTM 👍

PR #81, opened 2023-06-11
Introduction of Custom Kubernetes Operator for Enhanced Subsite Management

ddvdozuki:
# Problem
Our current approach to managing subsites has faced numerous challenges, particularly since they are created from within the app, yet carry significant infrastructure implications. Past solutions have fallen short of addressing these comprehensively, necessitating a robust solution to the following key issues:

1. **Ingress**: Our reliance on ingress controllers requires a dynamic solution to direct traffic to subsites as they are created by users.
2. **SSL**: The absence of wildcard certificate support for external domains in LetsEncrypt complicates SSL certification for subsite subdomains, demanding a separate certificate for each subsite.
3. **Grafana**: While we support automatic Grafana, Dashboard, and Datasource deployment for the primary (onprem) site, subsites go unnoticed by Grafana since the datasource is restricted to the primary database only.
4. **Feature Flags (subsite-level)**: Feature flags add complexity to the process, especially when a subsite wants different features than the main site. This introduces complications for an automated solution as it involves more than a simple row update. (Note: subsite-level feature flags are not addressed in this PR)

# Solution
We propose a custom Kubernetes operator: [dozuki-operator](https://github.com/Dozuki/dozuki-operator). This Go-based operator can efficiently handle the infrastructure logic necessary for managing subsites and feature flags. The creation of a new subsite triggers the following steps:

1. A sidecar container on the app pod detects a new row in the site index table and creates a new Subsite custom resource.
2. The operator recognizes the new resource and, based on its values, generates the required resources:
   * New ingress and SSL cert
   * A new private Grafana instance with pre-configured dashboards and datasource
   * The primary Grafana instance is updated to include the new dashboards and datasources, enabling site-wide data access in one place.
Additionally, the operator also handles the teardown when a subsite is disabled.

### Transition of Grafana to KOTS Manifests
To support these changes, Grafana's deployment needs to transition to KOTS manifests. The introduction of the Grafana operator also necessitates custom resources for Grafana, GrafanaDatasource, GrafanaDashboard, and GrafanaFolder to allow the dynamic creation of new Grafana instances. This change impacts the infrastructure related to secrets, configmaps, and jobs previously essential for Grafana, requiring deletion of old ones and addition of new ones that can be passed through to the initial KOTS configuration.

### KOTS Version Lock
Given our increased integration with specific KOTS features, such as Config values, it is recommended to lock the KOTS version instead of installing the latest version. This version locking can prevent potential issues during functionality testing.

## QA
This branch is deployable on any of our organizational AWS accounts. For deployment testing on GovCloud, ensure to set the external_fqdn variable to the CNAME record manually created.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,13 @@ jobs:
           tag_name: ${{ github.ref }}
           release_name: Release ${{ github.ref }}
           body: |
-            S3 Migration Support Fix
-            - Ignore bucket name and prefix changes on stack upgrade to prevent re-create.
+            App Ingress, LetsEncrypt certs, & Monitoring Overhaul
+            - We now use ingress for the admin dashboard, main app, and grafana (at /dashboards).
+            - All SSL certs are managed automatically via cert-manager and LetsEncrypt.
+            - New DNS subdomains auto-provisioning for dozuki.cloud.
+            - Grafana refactor to use MySQL for settings and user persistence.
+            - Slack & email notifications for monitoring alarms.
+            - New monitoring alarms for EKS.
+            - RDS Read Replica for BI now available and set as default over DMS.
           draft: false
           prerelease: false


jamesmattisondozuki:
Looks reasonable to me

caphene (approved):
Nice work 👍

terraform/physical/eks.tf
```diff
@@ -142,8 +142,12 @@ data "aws_iam_policy_document" "eks_worker" {
       "s3:GetObject",
       "s3:GetObjectAcl",
       "s3:ListBucket",
-      "s3:ListObjectsV2",
```
scbarber, line 145: Are we sure we don't need `ListObjectsV2`?

scbarber (commented):
Generally looks good, but I have a question about an S3 action that was removed.

scbarber (approved):

PR #83, opened 2023-06-11
Enhancements to Monitoring Infrastructure via RDS Alarms and Necessary Refactoring

ddvdozuki:
In light of our expanding customer base and growing utilization of PC, it has become increasingly crucial to augment our existing monitoring capabilities, currently centered around application and EKS instance level alarms. Our long-term aim is to integrate Prometheus comprehensively, however, as an immediate step, I have introduced CloudWatch alarms for the RDS instances.

The complexity involved in this task was substantial, particularly concerning the creation of the "free memory" alarm. RDS and CloudWatch lack metrics that indicate the **percentage** of free memory, only offering an absolute free memory metric. This poses a challenge, considering the variability in our RDS instance types. To circumnavigate this issue, I had to resort to hard-coding an array of potential RDS instance types, complete with their respective memory allocations, and incorporating this into the alarm as a metric query. Although hard coding such data into the Terraform isn't ideal, it became necessary due to the lack of an alternative means to obtain this information from Amazon without invoking a raw API call, necessitating a null provisioner. This method, though not perfect, was chosen as the lesser of two evils.

In addition to this, I've also initiated the use of `aws_rds_orderable_db_instance` to select the first instance from the list of preferred instances that satisfies our specifications. This method also verifies if the selected instance type corresponds to one from our hard-coded list, using the postcondition lifecycle. This allows us to ensure compatibility with our current setup. However, if we decide to support instances beyond this list, we can reassess this methodology in the future.

I believe these changes are a significant stride towards improving our infrastructure's reliability, especially as we continue to scale. I appreciate your understanding of the complexities involved and invite your feedback on these changes.

### DIFF
--- a/terraform/physical/README.md
+++ b/terraform/physical/README.md
@@ -39,7 +39,13 @@
 | <a name="module_nodes_alarm"></a> [nodes\_alarm](#module\_nodes\_alarm) | terraform-aws-modules/cloudwatch/aws//modules/metric-alarm | 4.2.1 |
 | <a name="module_primary_database"></a> [primary\_database](#module\_primary\_database) | terraform-aws-modules/rds/aws | 5.6.0 |
 | <a name="module_primary_database_sg"></a> [primary\_database\_sg](#module\_primary\_database\_sg) | terraform-aws-modules/security-group/aws | 4.17.1 |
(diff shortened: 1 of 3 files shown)


caphene (approved):
Good work 👍

scbarber (approved):

PR #88, opened 2023-07-03
Refactor Logical Module to Remove Dozuki Configmap and Other Appropriate Resources

ddvdozuki:
This PR moves many resources from terraform to KOTS (see companion PR: https://github.com/Dozuki/kots-app/pull/13). I'll copy the PR message here since that repo is private:

> This Pull Request aims to address long-standing challenges associated with race conditions and operational sequencing, which have been occurring between the Terraform "logical" module and the Replicated release. The solution proposed involves transitioning several resources, previously deployed via Terraform, to be managed directly by Kubernetes manifests under the Replicated (KOTS) framework. Here are the resources involved in this transition:
> 
> 1. Dozuki Configmap: This app configuration, which gets mounted to /etc/dozuki on the app pod.
> 2. Frontegg Helm Chart: The Helm chart that manages the installation of Frontegg and all its related subcharts.
> 3. Frontegg DB Creation: The Kubernetes job assigned with the task of generating empty databases for Frontegg.
> 4. MongoDB/Redis Helm Chart: These are service dependencies for Frontegg.
> 
> To enable this transition, we have migrated much of the overarching infrastructure configuration into Replicated via Config. Terraform now relays all the values it used to employ for creating the app configmap and the replaced resources to KOTS during installation.
> 
> The motivation for this transition is twofold. Firstly, we have often discussed the potential of substituting Replicated/KOTS in the future, which led us to consciously avoid migrating resources into the manifests, and thus most on-stack resources were created using Terraform. However, after thorough consideration and exploration, it became apparent that even if we decided to replace KOTS with a different system (which is a considerable "if"), that new system would still need to manage these resources. Hence, Terraform would not be a suitable platform for deploying these resources regardless of the scenario.
> 
> Additionally, after further research, it appears that migrating the remaining resources from the Terraform logical module into our app deployment layer could potentially be a more suitable approach. This change would allow us to better streamline our deployment process and mitigate the issues that have previously arisen with our current setup.
>

### DIFF
--- a/.github/workflows/terraform.yml
+++ b/.github/workflows/terraform.yml
@@ -46,7 +46,7 @@ jobs:
         github_token: ${{ secrets.github_token }}
         working_directory: "./terraform/${{ matrix.module }}" # Optional. Change working directory
         reporter: github-pr-review # Optional. Change reporter
-        fail_on_error: "true" # Optional. Fail action if errors are found
+        fail_on_error: "false" # Optional. Fail action if errors are found
         filter_mode: "nofilter" # Optional. Check all files, not just the diff
         tflint_rulesets: "aws" # Optional. Extra official rulesets to install
 


jamesmattisondozuki (approved):
LGTM

scbarber (approved):

caphene (approved):

PR #91, opened 2023-07-11
Enhancing DMS Performance: Updates to Replication Setup and Schema Mapping Strategy

ddvdozuki:
In our ongoing efforts to minimize the failure rate of DMS, we have implemented several alterations to our conditional replication setup and our approach to establishing and connecting to the replica database. Most importantly, we now completely turn off multi-az for the target replica, aligning with AWS' best-practice recommendation. Likewise, we have disabled foreign key checks in the target database to further enhance performance and reliability.

The most significant shift, however, is in our strategy for schema mapping. Rather than solely relying on an exclusive blacklist, we've transitioned to a dual-pronged approach that incorporates both a whitelist and a blacklist as needed. This shift aims to better manage and mitigate any potential errors by excluding databases and tables that should remain untouched or may pose an increased risk of replication errors.

### DIFF
--- a/terraform/physical/bi.tf
+++ b/terraform/physical/bi.tf
@@ -119,7 +119,7 @@ resource "aws_dms_endpoint" "target" {
   ssl_mode                    = "verify-full"
   endpoint_type               = "target"
   engine_name                 = "mysql"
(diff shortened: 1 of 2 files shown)


jamesmattisondozuki (approved):
Looks fine to me... note that there is a warning about one of the files regarding key rotation.

PR #92, opened 2023-07-11
Correction to KOTS Configuration Update: Preventing Overwrites with --merge Tag

ddvdozuki:
In the course of incorporating the feature to update the KOTS configuration each time the terraform reruns the logical module, I mistakenly caused the entire configuration to be overwritten. This error stemmed from an assumption that the default behavior would be a "merge", but this turned out to be incorrect. To rectify this, it is necessary to explicitly use the --merge tag. Doing so will prevent the function from inadvertently erasing any customized settings made from the dashboard.


jamesmattisondozuki (approved):
Looks good. Gonna just merge this in.

PR #93, opened 2023-07-15
Automated Recovery for DMS Task Failures and Minor Refactoring

ddvdozuki:
Due to persistent issues encountered with AWS Database Migration Service (DMS) relating to our MySQL views, we have devised an alternate solution to improve reliability.

This PR introduces a new AWS Lambda function, designed to monitor the state of DMS replication tasks. When a DMS task encounters an error and changes its state to "ERROR", this Lambda function will automatically restart the task, while still allowing for regular manual interruptions.

In addition to this central update, this PR includes a handful of minor refactoring changes:

1. **Isolated Run Script Enhancements:** Modifications have been made to our isolated run script for improved reliability. We have moved away from using `git worktree` for repo duplication due to certain challenges it presented in copying the current branch. Instead, a straightforward rsync approach was found to be more effective. Furthermore, the Docker tag for the isolated run container is now locked to a specific version of the source container.

2. **DMS Role Relocation:** The DMS role and its associated permissions have been removed from the bootstrap module and are now managed by the main module.

3. **Extended EKS Permissions:** To improve cluster management, we have added EKS permissions to allow any user with the AWSAdministrator role to manage the cluster. This important enhancement was overlooked previously, but now allows us to configure the cluster context without assuming a third-party role.


caphene (approved):
Nice work 👍

PR #125, opened 2026-03-05
Restore frontegg database creation job for webhooks environments

ddvdozuki:
## Summary
- Restore `frontegg_events` and `frontegg_webhooks` database creation job that was removed in fd5809a when frontegg was moved to KOTS
- The connectivity subchart services (event-service, webhook-service) require these databases to exist before they can start
- Job is gated on `enable_webhooks` and runs after the helm release is installed

## Changes since initial review
- **Replaced** `imega/mysql-client:10.6.4` with hardened MySQL image per review feedback
- **Improved** credential handling: DB host/user/password stored in a Kubernetes Secret and injected via `secretKeyRef` instead of inline CLI arguments
- **Updated** `kubernetes_config_map` to `kubernetes_config_map_v1`

## Test plan
- [x] Deployed to ddv-test hooks — job completed, both databases created
- [x] event-service and webhook-service started successfully after database creation
- [ ] Verify hardened MySQL image pulls successfully and mysql client works with Aurora


ddvdozuki:
Superseded by #129 which includes the updated webhooks.tf with hardened MySQL image and secret-based credential handling.

PR #136, opened 2026-06-02
Fix CloudWatch alarms stuck in INSUFFICIENT_DATA

ddvdozuki:
## Problem

Most CloudWatch alarms on a customer stack sit in **INSUFFICIENT_DATA** despite the instances running for weeks. Investigation found **two independent root causes**, both affecting any current/Auto Mode deploy (older stacks predate them and are fine).

### 1. RDS alarms — wrong metric dimension

All `AWS/RDS` alarms set `DBInstanceIdentifier = module.primary_database.db_instance_id`. A live check showed the dimension resolving to the RDS **resource ID** (`db-XXXXXXXX…`) instead of the instance identifier, while correctly-working stacks show the plain identifier.

The rds module is pinned to **v5.6.0**, whose `db_instance_id` output is `aws_db_instance.this.id`. Under **AWS provider v5+**, `aws_db_instance.id` returns the resource ID (`db-XXXX`), not the instance identifier. CloudWatch's `AWS/RDS` namespace keys on the identifier, so the dimension matches no published metric → every RDS alarm is stuck. Stacks applied under provider v4 (where `.id` was the identifier) are unaffected.

**Fix:** point all 7 RDS alarm dimensions at `local.identifier` — the value already passed to the module as `identifier =`, so it *is* the DB identifier by construction, and it's immune to provider/module-version drift. (Bumping the module isn't viable: latest needs Terraform ≥1.11.1 vs our Spacelift-pinned 1.5.7, and newer versions removed the `db_instance_id` output anyway.)

### 2. Container Insights alarms — agent can't authenticate

`*-cpu-high`, `*-memory-utilization`, `*-out-of-disk` read the `ContainerInsights` namespace, which has **0** metrics for the new cluster (a working cluster has hundreds). The `amazon-cloudwatch-observability` addon is installed and its pods run, but it has no `serviceAccountRoleArn`/pod-identity. EKS Auto Mode gives nodes a minimal role, so the agent falls back to it and every publish is denied:

```
AccessDeniedException: ...assumed-role/<cluster>-eks-auto-.../i-...
  is not authorized to perform: logs:PutLogEvents / xray:PutTraceSegments
```

The node role carries only `AmazonEKSWorkerNodeMinimalPolicy` + `AmazonEC2ContainerRegistryPullOnly`. Auto Mode intentionally keeps it minimal and expects workloads to get permissions via Pod Identity. The addon was also installed **out-of-band** (not in Terraform).

**Fix:** manage the addon in `eks.tf` with an IAM role (`CloudWatchAgentServerPolicy` + `AWSXRayDaemonWriteAccess`) bound to the `cloudwatch-agent` SA via `pod_identity_association`. Both the metrics agent and fluent-bit run as that SA, so one association covers metrics and logs.

## Changes
- `terraform/physical/monitoring.tf` — 7 RDS alarm dimensions → `local.identifier` (+ explanatory comment)
- `terraform/physical/eks.tf` — `amazon-cloudwatch-observability` addon + IAM role + pod-identity association
- `terraform/physical/README.md` — regenerated

## ⚠️ One-time step for any stack that already has the addon
A stack with the addon installed manually will hit `ResourceInUseException` on Terraform's `CreateAddon` at first apply. Before applying, remove the orphaned addon once so Terraform creates the managed one:

```
aws eks delete-addon --cluster-name <cluster> --addon-name amazon-cloudwatch-observability --region <region>
```

(Safe — the orphan publishes nothing today.) New stacks have no addon and apply cleanly. RDS alarm changes are dimension-only (no resource replacement).

## Validation
- `terraform fmt` clean; `terraform validate`/tflint pass (the only `validate` errors are a pre-existing unrelated `dns` aliased-provider/state artifact from validating outside the Terragrunt wrapper)
- Root causes confirmed live: RDS alarm dimension = resource ID; ContainerInsights namespace empty; agent logs show `AccessDenied` against the Auto Mode node role; node role lacks CloudWatch policies

### DIFF
--- a/terraform/physical/README.md
+++ b/terraform/physical/README.md
@@ -61,6 +61,7 @@
 | [aws_dms_replication_instance.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dms_replication_instance) | resource |
 | [aws_dms_replication_subnet_group.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dms_replication_subnet_group) | resource |
 | [aws_dms_replication_task.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dms_replication_task) | resource |
+| [aws_eks_addon.cloudwatch_observability](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_addon) | resource |
 | [aws_eks_pod_identity_association.app_default](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_pod_identity_association) | resource |
 | [aws_eks_pod_identity_association.app_migration_wait](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_pod_identity_association) | resource |
 | [aws_eks_pod_identity_association.cert_manager](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_pod_identity_association) | resource |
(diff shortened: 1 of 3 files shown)


ddvdozuki:
Superseded by #137 (recreated to remove stray commit history).

PR #217, opened 2026-06-24
refactor(logical): use the dozuki chart's seaweedfs subchart on Azure

ddvdozuki:
## Summary

On Azure the dozuki app needs an S3-compatible object store (Azure has no S3). Until now CPI ran its own **standalone** SeaweedFS install (an upstream `seaweedfs` helm release plus an aws-cli bucket-creation Job) in `terraform/logical/seaweedfs.tf`. This PR removes that and instead enables the **dozuki chart's bundled seaweedfs subchart** (`seaweedfs.enabled=true`) — exactly the way an on-prem install provisions SeaweedFS. AWS (real S3) and on-prem (already on the subchart) are unchanged.

## Before / after (Azure only)

**Before**
- `helm_release.seaweedfs` — standalone upstream seaweedfs chart (4.33.0): master/volume/filer with the embedded S3 gateway, leveldb filer store on a PVC, `s3.credentials.admin.*` from `random_password`.
- `kubernetes_job_v1.seaweedfs_buckets` — aws-cli Job creating the four buckets against the seaweedfs S3 endpoint.
- `helm_release.app` set `objectStorage.endpoint`/`publicHost` to `s3.<domain>`, `objectStorage.credentials.*` from the random passwords, and `depends_on` the two resources above.

**After**
- The dozuki chart's seaweedfs subchart owns the master/volume/filer StatefulSets, the embedded S3 gateway Service (`dozuki-seaweedfs-filer:8333`), the **MySQL** filer metadata store, schema init, and bucket creation (the subchart's native `filer.s3.createBuckets` hook — no aws-cli, airgap-friendly).
- `helm_release.app` enables it via `local.seaweedfs_values` (merged into the existing Azure values), and the app's `objectStorage.endpoint`/`credentials` point at the embedded filer S3 service.

## What was removed
- `helm_release.seaweedfs` (standalone upstream chart).
- `kubernetes_job_v1.seaweedfs_buckets` (aws-cli bucket Job) — replaced by the subchart's `createBuckets` hook.
- The two `depends_on` entries for those resources in `helm_release.app`.
- The old `local.seaweedfs_s3_endpoint` value (`seaweedfs-s3...:8333`) — replaced with the chart's filer S3 service name.

(`random_password.seaweedfs_access_key` / `seaweedfs_secret_key` are kept and relocated into the rewritten `seaweedfs.tf`; a new `random_password.seaweedfs_filer_db` was added for the filer's dedicated MySQL user.)

## Standalone config → chart `seaweedfs:` values mapping
| Old standalone | New chart subchart value |
|---|---|
| `s3.credentials.admin.accessKey/secretKey` | `seaweedfs.s3.accessKey/secretKey` (rendered into `dozuki-seaweedfs-s3` secret) |
| leveldb filer on PVC | `seaweedfs.mysqlFiler.*` (MySQL metadata; on-prem model). Reuses the app DB server; `adminUsername/adminPassword` = app DB creds for the schema-init Job; dedicated `seaweedfs` user + `seaweedfs_filer` schema |
| `master/volume/filer ... storageClass: managed-csi` | `seaweedfs.{master,filer}.data.storageClass` + `seaweedfs.volume.dataDirs[].storageClass` = `managed-csi` |
| `volume.dataDirs[].size = <var>Gi` | `seaweedfs.volume.dataDirs[].size` (same `var.seaweedfs_volume_size_gb`) |
| `defaultReplication 001`, `volume.replicas 2` | `global.seaweedfs.enableReplication=true` + `replicationPlacement="001"`, `seaweedfs.volume.replicas=2` |
| aws-cli bucket Job over `var.s3_*_bucket` | `seaweedfs.filer.s3.createBuckets` driven from the **same** `var.s3_*_bucket` list |
| `objectStorage.endpoint/publicHost = s3.<domain>` | endpoint now the in-cluster filer S3 service; `publicHost` kept for external presigned access |

## Things that did not map cleanly — please review
1. **Bucket creation.** The chart does NOT use an aws-cli Job; it uses the subchart's native `filer.s3.createBuckets` hook (SeaweedFS image, airgap-safe). The chart's packaged default is the FIXED list `images/pdfs/docs/objects` and the chart docs say `objectStorage.*Bucket` must stay at those defaults when seaweedfs is enabled. CPI historically uses per-install bucket names (`var.s3_*_bucket`), so I **override** `createBuckets` from those same vars — the created buckets always equal `objectStorage.*Bucket`, preserving CPI's naming. Helm replaces lists (not append), so the chart's fixed list is fully replaced. Confirm we're OK keeping CPI's custom bucket names rather than switching Azure to the chart's `images/pdfs/docs/objects`.
2. **MySQL filer is new for Azure.** The standalone filer used leveldb on a PVC; the chart subchart requires MySQL (matching on-prem). It now provisions a `seaweedfs_filer` schema + `seaweedfs` user on the **app database server** via the chart's schema-init Job (run as the app DB admin). This adds a dependency on the app DB being reachable/admin-capable at install. Please confirm the Azure app DB admin can `CREATE DATABASE/USER/GRANT`.
3. **Public S3 backend service name.** The chart's `objectStorage.publicBackend.service` default is `seaweedfs-s3`, which only matches a standalone S3 deployment. We run S3 **embedded in the filer**, so I override `publicBackend.service` to `dozuki-seaweedfs-filer` (port 8333) so the public S3 HTTPRoute (`objectStorage.publicHost`) resolves.
4. **`global` shallow-merge.** Terraform `merge()` is shallow, so the subchart replication settings live under the existing `global` block in `kubernetes.tf` (alongside `imagePullSecrets`) rather than in `local.seaweedfs_values`, to avoid clobbering the GHCR pull secret.

## Validation
- `terraform fmt` clean.
- `tofu validate` passes for these changes. The only error is the pre-existing `provider "vault"` "Missing required argument: address" (address comes from `VAULT_ADDR` at runtime) — present on master, unrelated.
- `terraform-docs` / `tflint` pre-commit hooks fail repo-wide on the pre-existing `azurerm.main[each.key]` `for_each` provider syntax and a pre-existing `main.tf:299` expression; the logical `README.md` resource table was updated by hand instead.

Source-of-truth verified against the bundled subchart `seaweedfs-4.31.0` and the chart's `seaweedfs/` templates (filer Service exposes `swfs-s3`/8333; `filer.s3.existingConfigSecret` + `createBuckets`; `dozuki.objectStorageEndpoint` helper).

### DIFF
--- a/terraform/logical/README.md
+++ b/terraform/logical/README.md
@@ -41,7 +41,6 @@ No modules.
 | [helm_release.envoy_gateway](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release) | resource |
 | [helm_release.external_dns](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release) | resource |
 | [helm_release.external_secrets](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release) | resource |
-| [helm_release.seaweedfs](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release) | resource |
 | [kubernetes_cluster_role_binding_v1.dozuki_list_role_binding](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/cluster_role_binding_v1) | resource |
 | [kubernetes_cluster_role_binding_v1.vault_auth_delegator](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/cluster_role_binding_v1) | resource |
 | [kubernetes_cluster_role_v1.dozuki_list_role](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/cluster_role_v1) | resource |
@@ -52,7 +51,6 @@ No modules.
 | [kubernetes_job_v1.dms_start](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/job_v1) | resource |
 | [kubernetes_job_v1.frontegg_db_create](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/job_v1) | resource |
 | [kubernetes_job_v1.grafana_db_create](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/job_v1) | resource |
-| [kubernetes_job_v1.seaweedfs_buckets](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/job_v1) | resource |
 | [kubernetes_manifest.nodepool_on_demand](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest) | resource |
 | [kubernetes_manifest.nodepool_spot](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest) | resource |
 | [kubernetes_manifest.tgb_http](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest) | resource |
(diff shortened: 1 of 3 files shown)


ddvdozuki:
Addressed review feedback on bucket naming (commit 4ce543f):

SeaweedFS bucket names are now **static** (the chart's defaults), not the per-install `var.s3_*_bucket`.

- Dropped the `createBuckets` override + `local.seaweedfs_buckets`; the subchart's native hook creates the chart-default set: **`images`, `pdfs`, `docs`, `objects`** (note: documents bucket is `docs`).
- The app's `objectStorage.imagesBucket/pdfsBucket/documentsBucket/objectsBucket` are now per-cloud: **azure → the static chart name**, **aws → `var.s3_*_bucket`** (real S3). The chart does not auto-wire these (the app reads `objectStorage.*Bucket` verbatim), so they're set explicitly. App ↔ SeaweedFS bucket names match exactly: images/pdfs/docs/objects.
- Added comments at both the `var.s3_*_bucket` usage (kubernetes.tf) and in seaweedfs.tf stating `var.s3_*_bucket` is real-S3/AWS only and does not apply to SeaweedFS/Azure.

`terraform fmt` clean; `tofu validate` shows only the pre-existing `provider "vault"` address error.

PR #262, opened 2026-07-23
feat(logical): istio ambient mesh behind istio_mesh_state

ddvdozuki:
Adds Istio ambient (1.30.3) to the logical layer behind a single lifecycle enum,
default disabled. States: disabled, installed (base/istiod/cni/ztunnel in
istio-system plus Karpenter startup taints and the untaint controller),
permissive (dozuki, envoy-gateway-system, redis-system enrolled; ratelimit redis
NetworkPolicy gains the HBONE port; ztunnel PodMonitor), strict (namespace-wide
STRICT PeerAuthentication with carve-outs for the NLB-facing envoy ports, the
envoy-gateway topology-injector webhook, the ESO and kube-prometheus-stack
admission webhooks, and the prometheus-adapter APIService callback).

Istio CRDs are owned by the base chart (templated, Helm-upgradable), unlike the
envoy-gateway vendored set. Gateway API CRDs are shared with envoy-gateway, so
istiod depends on that bundle. external-secrets is now version-pinned; the
strict carve-outs target its rendered webhook.

The carve-out list has five entries: the NLB-facing envoy proxy ports, the
envoy-gateway topology-injector webhook (controller pods, port 9443,
failurePolicy Ignore - without the carve-out, STRICT mTLS silently breaks
zone-aware routing since the webhook call from the API server fails open but
the routing behavior it configures never lands), the ESO admission webhook,
the kube-prometheus-stack admission webhook, and the prometheus-adapter
APIService callback. Carve-outs apply before namespace-wide STRICT in both
directions: rolling forward, the labels land, then the carve-outs, then STRICT;
rolling back, STRICT lifts before the carve-outs are removed.

Enabling the mesh (moving off disabled, to installed and beyond) adds
startupTaints to the spot and on-demand NodePools. Karpenter treats a NodePool
spec change as drift, so the first enable rolls every node on those custom
pools, and a later walk-back to disabled rolls them again removing the taint.
Envs should schedule the flip like a node-replacement maintenance event, not a
routine config change.

Plan guards: non-disabled states require commercial AWS (gov waits on the image
mirror, azure unsupported).

Verification: a fresh-env harness plan at the default (disabled) state shows
zero mesh resources among the logical layer's resources. Walking the state
enum step by step matches the design: installed adds 5 resources (namespace +
4 releases), permissive adds 3 (the enrollment labels) plus the NetworkPolicy
HBONE port and the ztunnel PodMonitor, strict adds 8 (3 namespace-wide STRICT
PeerAuthentications + 5 carve-outs). The bogus-value guard fails validation as
designed, with zero mesh resources leaking through even in that broken plan.
Full per-state resource lists and the guard error are in a PR comment. A true
no-op ("no changes") plan against a real previously-applied env wasn't
available in this pass; that lands with the next repin on an env that already
has the layer applied.

Note on external-secrets: the chart was already pinned at 2.8.0 on master. An
env like dev-min currently runs 2.6.0, so its next apply upgrades ESO
regardless of this branch. The strict-mode carve-out targets the rendered
webhook contract (labels, namespace, port), which is identical across both
versions.

Lifecycle: istio_mesh_state walks disabled -> installed -> permissive ->
strict one step per apply in either direction; skipping a step isn't
supported.


ddvdozuki:
## State-walk plan evidence (fresh-env harness, cluster `dev-min`)

Ran the logical layer plan at each `istio_mesh_state` value from a fresh (never-applied)
state. No mesh resource ever appears at `disabled`; each step forward adds exactly the
resources gated on that state, no extras.

### disabled (default)
Zero mesh resources among the layer's resources. No `istio_system` namespace, no
`istio_*`/`ztunnel` releases, no `ambient_*` labels, no `peer_auth_*` manifests anywhere in
the plan.

### installed (+5 over disabled)
```
kubernetes_namespace_v1.istio_system[0]
helm_release.istio_base[0]
helm_release.istiod[0]
helm_release.istio_cni[0]
helm_release.ztunnel[0]
```
NodePool `startupTaints` (`cni.istio.io/not-ready`, `NoSchedule`) also appear on the spot/
on-demand NodePools at this step (absent at disabled).

### permissive (+3 over installed, plus the NetworkPolicy port and the PodMonitor)
```
kubernetes_labels.ambient_dozuki[0]
kubernetes_labels.ambient_envoy_gateway[0]
kubernetes_labels.ambient_redis[0]
```
The ratelimit redis NetworkPolicy gains the HBONE `15008` ingress port at this step. The
ztunnel PodMonitor (`kubectl_manifest.ztunnel_podmonitor`) is also gated at this state; it
didn't render in this particular harness run because of an unrelated blocker upstream in
the plan graph, not an istio gating issue.

### strict (+8 over permissive)
```
kubectl_manifest.peer_auth_strict["dozuki"]
kubectl_manifest.peer_auth_strict["envoy-gateway-system"]
kubectl_manifest.peer_auth_strict["redis-system"]
kubectl_manifest.peer_auth_carveouts["envoy-gateway-proxy"]
kubectl_manifest.peer_auth_carveouts["envoy-gateway-controller-webhook"]
kubectl_manifest.peer_auth_carveouts["kps-operator-webhook"]
kubectl_manifest.peer_auth_carveouts["external-secrets-webhook"]
kubectl_manifest.peer_auth_carveouts["prometheus-adapter"]
```
3 namespace-wide STRICT PeerAuthentications (dozuki, envoy-gateway-system, redis-system) +
5 port-level carve-outs.

### guard (bogus value)
```
Error: Invalid value for variable

  on variables.tf line 58:
  58: variable "istio_mesh_state" {
    ├────────────────
    │ var.istio_mesh_state is "bogus"

istio_mesh_state must be one of: disabled, installed, permissive, strict.

This was checked by the validation rule at variables.tf:63,3-13.
```
Zero mesh resources appear even in the partial plan OpenTofu still computes around this
error.

ddvdozuki:
Pilot results from dev-min (branch SHA pinned via Spacelift, full state walk disabled -> installed -> permissive -> strict):

- disabled: apply on a live env showed zero mesh resources (0 added / 2 changed, both expected: the ESO pin catching up and an app release update from the version delta).
- installed: 5 resources created in the documented order, both NodePools gained the startup taint, the untaint controller cleared it on every fresh node, and the custom-pool roll completed with the app healthy throughout. The cluster SG's self-referencing rule covers 15008/15012/15017 with no physical changes.
- permissive: live enrollment with no restarts; mutual_tls connections observed via the ztunnel PodMonitor; destination-side non-mTLS inbound matched the documented carve-out classes exactly.
- strict: the walk caught a real bug. The two Terraform-managed namespaces stripped the enrollment label on the next apply (plan showed the label going to null), silently un-enrolling dozuki and redis-system while the Helm-created envoy-gateway-system stayed enrolled. The negative test (plaintext from an unenrolled namespace) passing traffic exposed it. Fixed in 9cdd36a with targeted ignore_changes on that label key; after the fix, all enrollable workloads show HBONE (only hostNetwork node-exporters remain TCP, as designed), plaintext from outside the mesh is refused on both the service and direct pod-IP paths, HPA external metrics and webhook callbacks work through their carve-outs, rate limiting is healthy, and the public HTTPS path serves normally.

Operational notes for rollout: the installed flip drift-replaces every custom-pool node (schedule it as maintenance), and one pooled redis connection took a single reconnect at re-enrollment (expected RST-on-enrollment behavior).

ddvdozuki:
Load-test results from min (the loadtest/ k6 harness, which this branch also repairs - four latent bugs fixed: the seed category endpoint and guide type, seed pool persistence, a K6_VUS name collision that silently replaced the ramping scenario with a basic executor, and the guide page URL segment order; the harness had never actually run its designed profile before).

A/B at 50 VUs for 5m on identical infrastructure, flipping only istio_mesh_state between installed (no enrollment) and strict, client outside the mesh through the public path:

- p95: 140ms -> 172ms (+32ms). Throughput flat (22.5k vs 22.1k requests).
- Error rates 16.1% vs 17.6%, both sides dominated by the gateway's own per-IP rate limiting (the load generator sits behind one NAT IP), which affects the comparison equally.
- ztunnel during the run: under 200 millicores per node, single-digit MiB.

Squeeze at 300 VUs strict: ~240 rps offered, two thirds shed by rate limiting as designed, ztunnel peaked at 514 millicores and 20MiB, zero ztunnel restarts, istiod stable, application healthy, and strict enforcement verified intact afterward.

The state walk for the baseline also exercised staged rollback end to end (strict -> permissive -> installed and back), including recovering from a Spacelift runner flake mid-walk. One follow-up: the harness summary's p99 extraction prints 0 and needs a fix.

ddvdozuki:
Chaos / hammer battery on min at strict mTLS, every experiment under continuous in-cluster traffic. All passed.

- ztunnel kill (traffic node): replacement ready 5s, zero failed requests. The in-pod redirect rules survive the agent, so the gap is latency, not errors.
- istiod outage: held at 0 replicas ~3.5 min, data plane never dropped a request. A node provisioned during the outage stalled on the startup taint as designed; on restore istiod reached 2/2 in ~37s and the untaint controller cleared it immediately. Both resilience and the documented capacity-stall failure mode confirmed.
- Node delete under load: Karpenter replaced capacity, ~12 fresh nodes cycled through the taint and untainted within ~30s, one failed request (a pod mid-eviction) across the whole event. The app HPA kept reading live external metrics through the prometheus-adapter carve-out.
- #60882 (ungraceful reboot): EKS Auto Mode denies ec2:RebootInstances on managed nodes, so the operator path to this bug does not exist. Reproduced the mechanism directly instead - wiped both dataplane agents while an app pod ran untouched through the ~80s gap; plaintext from outside the mesh stayed rejected throughout (fail-closed). No bypass.
- redis kill (enrolled ns): recovered 6s, gateway rate-limit fails open, no customer impact.
- beanstalkd kill (singleton, server-first protocol over HBONE): recovered ~23s, queueworkerd reconnected with zero restarts, post-recovery upload processed end to end. Validates server-first protocol under ztunnel.
- istiod rolling restart under load: all ztunnels reconnected, strict still enforcing, cert path re-exercised clean.
- Full app-tier rollout under load: every fresh pod re-enrolled HBONE, traffic clean.

Also validated end-to-end under strict: image and PDF upload (app -> beanstalkd -> queueworkerd -> S3, all cross-pod over the mesh). Note: min's allowed-document-extensions table was empty, so uploads were broken on min before the mesh ever existed - seeded it. Prometheus target sweep: 94/95 up.

Final state: istiod 2/2, cni+ztunnel 19/19, zero non-running app pods, 8 PeerAuthentications, negative test rejected. One fast-follow noted: a standing bypass-detection alert (plaintext between enrolled workloads), though every reproducible scenario was fail-closed.

PR #280, opened 2026-07-25
feat(logical): wire the operator's primary sitename

ddvdozuki:
## What

New `operator_primary_site` variable (default `"onprem"`), passed through as
`dozuki-operator.features.primarySite`.

## Why

The operator applies every entitled `FeatureRequest` to the **siteid-2** site plus each Subsite
CR (`features/sites.go` `TargetSites`). `PRIMARY_SITE` comes from
`charts/dozuki-operator/values.yaml`, which defaults it to **`onprem`** - and **nothing
overrode it**, on either the parent dozuki chart or this layer. So that default reached every
install.

It is only right for single-site CloudPrem. A platform seeded from a customer database has that
customer's sitename at siteid 2. When it is wrong, every reconcile fails:

```
SiteNotFoundException: Site 'onprem' not found
  SiteSpecific::getSiteid('onprem')
  <- SiteToolsCLI->updateSetting('onprem', 'feature-courses', 'true')
```

and it never self-heals. Found staging the shared GovCloud cutover, where siteid 2 is `dozuki`:
`onprem` sorted 5th of 11 and - because the operator's fan-out script used `set -e` - six of the
ten real tenants silently never received the flag. The `set -e` half is fixed in
**dozuki-operator #28**; this is the configuration half, and the two are independently worth
having.

## Safety

Default is `"onprem"`, i.e. exactly today's behaviour, so no existing stack changes. Consumers
that need it set it per env:

```hcl
operator_primary_site = "dozuki"
```

The variable description carries the query for finding the right value
(`SELECT name FROM sites.site_index WHERE siteid = 2`).


ddvdozuki:
Closing: wrong shape, and CPI does not need another variable.

Investigating further showed the operator **already reads the sites database itself** -
`subsiteDiscoveryRunnable` polls `site_index` and projects Subsite CRs into k8s (`cmd/main.go`
`syncSubsitesFromDB`). So the primary sitename should not be configuration at all; the operator
can derive it from `SELECT name FROM sites.site_index WHERE siteid = 2`.

It also showed the identity is currently expressed twice and inconsistently: discovery hardcodes
`WHERE name NOT IN ('base', 'onprem')` in SQL, while `TargetSites` uses `PRIMARY_SITE` defaulting
to `onprem`. Neither derives it from siteid, which is why sharedgov ends up with `dozuki` as a
Subsite CR *and* a phantom `onprem` primary.

Being fixed properly in dozuki-operator #28. The existing operator chart value
(`features.primarySite` -> `PRIMARY_SITE`) stays as an override for odd installs, so no CPI
variable is needed.

PR #281, opened 2026-07-25
feat(logical): enable the in-cluster site-FQDN redirect on meshed envs

ddvdozuki:
Sets `istio.fqdnRedirect.enabled` from `local.mesh_enrolled` so the chart renders the ServiceEntry that keeps in-cluster fetches of the app's own site FQDN off the public NLB (wkhtmltopdf PDF generation otherwise hairpins out to the internet-facing LB and back). Off on non-meshed envs.

Pairs with the chart's `proxyProtocol.optional` (accepts the header-less in-cluster connection) - Dozuki/helm#140. Requires a chart version carrying both.

Adjacent to the `istio.ztunnelMonitor` set line from #277 - trivial rebase if that lands first.

### DIFF
--- a/terraform/logical/kubernetes.tf
+++ b/terraform/logical/kubernetes.tf
@@ -679,6 +679,11 @@ resource "helm_release" "app" {
 
     # --- Monitoring ---
     { name = "monitoring.enabled", value = "true" },
+    # Redirect in-cluster fetches of the app's own site FQDN to the in-cluster gateway (chart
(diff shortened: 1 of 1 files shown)


ddvdozuki:
Not needed - the chart renders the ServiceEntry off a CRD-capability check (no flag/variable), so there's nothing for the logical layer to set. Superseded by the capability-gated approach in Dozuki/helm#140.

PR #358, opened 2026-07-29
feat(physical): DR dead-man detection and acknowledged paging

ddvdozuki:
P2 of the failover-automation design: automatic detection, human decision.

The dead-man chain has no primary-region dependency: the headless DR secondary emits AuroraGlobalDBReplicationLag into DR-region CloudWatch, and the alarm treats missing data as breaching - when the primary region dies the metric goes silent and the silence pages. Sustained 5-minute replication lag pages through the same alarm. 15 one-minute datapoints keep fresh-stack warmup and transient spikes quiet while still paging well inside the ~50min rebuild RTO.

Routing: a DR-region SNS topic, owned by operations_owner (defaults from managed_private_cloud: MPC stacks page Dozuki on-call, customer-operated stacks page the customer). dr_paging_endpoint takes an acknowledged paging integration (PagerDuty/Opsgenie SNS inbound URL); with none set the topic falls back to alarm_email and a check block warns on every plan that DR detection is not operationally enabled - email wakes nobody.

The alarm description is the page: stack identity, the split-brain warning, and the literal cpi-dr status/prepare/promote commands, so the on-call has the runbook prompt even if the incident took the wiki with it. Nothing here can trigger a promotion.

No-op for stacks without the Aurora DR secondary; new alarm+topic in the DR region for those with it.


ddvdozuki:
Rebased on #359 and switched the page text to the provider-derived DR region - var.dr_region is empty without an env.hcl entry (the #359 root cause), which would have rendered the page's cpi-dr commands with blank --dr-region values. Merge #359 first.

PR #364, opened 2026-07-29
fix(deps): commit provider lock files and adopt releases on a soak delay

ddvdozuki:
## Why

Nothing pinned provider versions. `.gitignore` line 2 is `.terraform*`, which swallows `.terraform.lock.hcl`, so **every init resolved the newest version the constraints allowed** — a bad release reached the whole fleet, Spacelift included, the moment it published.

aws **6.57.0** did exactly that on 2026-07-29: data-source reads failed across IAM, KMS, EC2 and STS with `302 UnknownError`, `SerializationException`, and signature errors. Plans died during refresh, so nothing applied. Two harness cycles were lost before the cause was found. Upstream is hashicorp/terraform-provider-aws#49170 (a response-handling regression in `aws-sdk-go-base`), and Spacelift raised their own status-page incident for it.

## What actually does the work: the lock file

Commits the lock files for `terraform/physical` and `terraform/logical`. Verified sufficient on its own: with the constraint left as plain `~> 6.0` — which *permits* 6.57.0 — init still installed 6.56.0 and planned clean.

It does not make us smart about which versions are good. It changes provider adoption from *silent, automatic and immediate at init* into *a pull request with CI*. A bad version can then only arrive somewhere we'd see it.

The `!= 6.57.0` exclusion on master stays regardless: stacks pinned to older refs have no lock file, and it also means no resolver can pick that build again.

## Soak length: optimised for not going stale, not for being current

Being current has no value to us. Being stuck on an ancient provider does, and that is the failure mode this is tuned against.

Renovate always proposes **the newest version that has already passed the soak**, marking anything newer as pending — it does not sit and wait for the latest release to age. Forward motion is therefore guaranteed no matter how long the window is, which makes a long window nearly free. Set to **30 days**: long enough that anything proposed has a real public track record, and the only cost is currency we don't want.

Being explicit about the limit: **a soak delays a bad release, it does not make it unreachable.** If the newest *aged* version happens to be a bad one, it still gets proposed — just with a month of public reports attached to it. The lock file, CI and a human merging are the controls; the delay only improves the odds.

Security advisories bypass the delay. Terraform **module** bumps stay disabled — they change resource shape and the rds v5.6.0 pin is deliberate (v6 dropped `db_instance_password`).

## Two guards

**Platform coverage.** A lock file whose hashes don't cover the runner's platform breaks init *everywhere* — Spacelift is linux_amd64, and a lock built by a plain local init on a Mac can carry darwin hashes only. The hashes aren't labelled by platform, so you can't verify by reading the file. The workflow asserts it the only honest way: complete the lock for linux_amd64 + darwin_arm64 and fail if tofu had anything to add.

**Staleness.** A monthly job that fails once the locked aws provider passes 180 days. It deliberately does *not* depend on Renovate — if the app is never installed, or its PRs pile up unmerged, the drift is otherwise silent, and silent drift is exactly how you end up somewhere you can't upgrade out of.

## Before merging

- **Renovate has to actually be installed** on this repo/org or `renovate.json` is inert. Nothing breaks either way, and the staleness job is the backstop that notices. If we'd rather not add it, Dependabot now ships a default 3-day cooldown with no config — weaker, and its Terraform lock file support is thinner, but it needs no app install.
- The platform job runs on this PR, so its first execution is visible here before it can block anyone.

## Worth doing next

Wiring the existing upgrade harness to plan against a proposed provider bump would turn "wait and hope someone else reports it" into "we find out in half an hour." That is what would make this genuinely safe rather than probabilistically safe.

## Not covered

Stacks pinned to existing refs stay exposed until they bump to a ref containing these files — the same limitation as the constraint fix, and the reason nothing in this PR helps a stack that runs today. infra-tf has its own `~> 6.0` modules wanting the same treatment.


ddvdozuki:
## Verified where the lock file actually has to live

The whole PR rests on "a lock file committed in the module source is honored when terragrunt runs it from `.terragrunt-cache`". That is not obvious — terragrunt's own model is to copy the lock file *back* next to `terragrunt.hcl` and have you commit it there — so I tested it with an isolated fixture: module locked at `null 3.2.1`, constraint widened to `~> 3.0` so only the lock can hold it back. If the lock is honored you get 3.2.1; if not you get 3.3.0.

| terragrunt | lock in module source (this PR) | lock in unit dir (terragrunt's model) |
|---|---|---|
| **0.99.4** (Spacelift) | **3.2.1 — honored** | 3.2.1 — honored |
| **0.53.0** (local harness) | 3.3.0 — **ignored** | 3.3.0 — **ignored, and the unit lock was overwritten with 3.3.0** |

**So this works on the path that matters.** Spacelift runs terragrunt 0.99.4, and there the committed module lock pins the provider exactly as intended.

**But it does not work in `live/tests`.** The harness runs terragrunt **0.53.0**, which ignores the lock and then rewrites it with whatever it resolved. Two consequences worth knowing before merging:

1. The harness will keep resolving the newest allowed provider, so it stays exposed to exactly the kind of release that started this (it is how cycles 38 and 39 died).
2. More importantly, the harness would no longer be validating the same provider set that production runs — which quietly weakens the thing we rely on it for.

That gap is pre-existing and separate from this PR: `CLAUDE.md` documents the fleet pairing as **OpenTofu 1.11.8 with Terragrunt 0.99.4**, and the local harness is four hundred-odd releases behind on terragrunt. Bumping it is not a one-liner — 0.99 renamed the `--terragrunt-*` flags the harness passes — so it wants its own change rather than being smuggled in here.

Nothing above changes the recommendation to keep the lock in the module source: `infra_version` is pinned per stack and physical/logical share it, so the provider set becomes part of the versioned, tested artifact and rolls out on the cadence we already use. Putting locks in the units instead would add a second, independent drift axis (module ref × provider set) across 36 stacks.

PR #365, opened 2026-07-29
chore(harness): migrate to the post-0.78 terragrunt CLI and current toolchain

ddvdozuki:
First of three steps toward running current Terragrunt/OpenTofu everywhere. This one is CPI-only and touches no live infrastructure.

## The harness was already broken, not just stale

Terragrunt dropped the `--terragrunt-*` prefix in **0.78.0** and **removed** the old flags in **0.85.0**, taking `run-all` with them. No deprecation grace period — they hard-fail.

I verified against real binaries rather than trusting release notes:

```
$ terragrunt-0.99.4 init --terragrunt-non-interactive
Error: flag provided but not defined: -terragrunt-non-interactive

$ terragrunt-0.99.4 run --all init --non-interactive
Succeeded 1
```

**0.99.4 is the version `infra-live/_spacelift/stacks.tf` pins for every Spacelift stack today.** So the harness's flags are already rejected by the fleet's own terragrunt. It ran locally only because this laptop happened to have **0.53.0** on PATH, and `upgrade-tests.yml` — which downloads 0.99.4 — cannot have been passing.

## Changes

| Old | New |
|---|---|
| `run-all apply --terragrunt-non-interactive` | `run --all apply --non-interactive` |
| `destroy --terragrunt-non-interactive` | `destroy --non-interactive` |
| `state list` | `run -- state list` |

That last one isn't cosmetic. Since **0.88.0** terragrunt no longer forwards unrecognised commands to tofu, and `state` is not one of its shortcuts, so diagnostics capture would die with "unknown command". `output` *is* a shortcut and is left alone.

Also migrated: `cleanup-orphans.sh`, `teardown-orphans.sh`, `live/clean.sh`, `live/isolated_run.sh`.

## Why this went unnoticed for so long

Nothing pinned the local binary — no tgenv/asdf config anywhere in the repo, and `DEPLOYMENT.md` just said `brew install opentofu terragrunt`. `run.sh` checked only that the binaries *existed*. It now asserts **>= 0.78** up front instead of failing twenty minutes into a run.

The floor is deliberately what the *code* needs, not what the fleet pins — gating on 1.1.1 would reject a perfectly working 0.99.4 setup for no reason.

## Version targets

CI moves to **OpenTofu 1.12.5** + **Terragrunt 1.1.1**, both confirmed present in Spacelift's bundled set (queried live via GraphQL).

These two move **together**: `stacks.tf` pins 1.11.8 specifically because Terragrunt 0.99.4 only pairs with 1.11.x, and 1.12.x requires Terragrunt >= 1.0.5. Moving to 1.1.1 is what unlocks 1.12.5. `terraform.yml` also goes 1.12.2 -> 1.12.5 so we stop validating on a different version than we apply with.

## Next steps (not in this PR)

1. **Validate**: a full `SCENARIO=recover` harness cycle on tofu 1.12.5 + terragrunt 1.1.1. Blocked on the currently-running cycle finishing, since `tgenv` switches the active binary globally.
2. **infra-live**: bump `stacks.tf` to 1.12.5 / 1.1.1. Fleet-wide, so it goes after validation. **Open question before that lands:** `terragrunt_version` applies to *all* stacks including the Vault ones on `TERRAFORM_FOSS` 1.5.7 — Terragrunt 1.1.1 against Terraform 1.5.7 needs confirming.

## Also noticed, deliberately not fixed here

- `terraform/logical/main.tf` declares `required_version >= 1.7.0`, which understates the real floor (OpenTofu >= 1.9 for `provider for_each`; docs say >= 1.11.1). Tightening it is a behavioural change for consumers and wants its own PR.
- `live/Dockerfile` pins `tf-1.5.2-tg-0.48.1` and cannot run the physical layer at all (`>= 1.11.1`). That container path looks abandoned — worth deleting or rebuilding rather than leaving as a trap.

### DIFF
--- a/.github/workflows/terraform.yml
+++ b/.github/workflows/terraform.yml
@@ -38,7 +38,10 @@ jobs:
       if: matrix.tool == 'tofu'
       uses: opentofu/setup-opentofu@v1
       with:
-        tofu_version: "1.12.2"
+        # Match what the fleet runs (infra-live/_spacelift/stacks.tf). Validating on a
+        # different version than Spacelift applies with is how a module change passes
+        # CI and then fails on a real stack.
+        tofu_version: "1.12.5"
 
     - name: Setup Terraform
       if: matrix.tool == 'terraform'
--- a/.github/workflows/upgrade-tests.yml
+++ b/.github/workflows/upgrade-tests.yml
@@ -46,12 +46,15 @@ jobs:
 
       - uses: opentofu/setup-opentofu@v1
         with:
-          tofu_version: "1.11.8" # highest 1.11.x in Spacelift's bundled set; pairs with Terragrunt 0.99.4, satisfies physical's >= 1.11.1
+          # Keep in step with infra-live/_spacelift/stacks.tf. The pair moves together:
+          # OpenTofu 1.12.x needs Terragrunt >= 1.0.5, which is why the fleet sat on
+          # 1.11.8 while it was on Terragrunt 0.99.4.
+          tofu_version: "1.12.5"
 
       - name: Install terragrunt
         run: |
           curl -fsSLo /usr/local/bin/terragrunt \
-            https://github.com/gruntwork-io/terragrunt/releases/download/v0.99.4/terragrunt_linux_amd64
+            https://github.com/gruntwork-io/terragrunt/releases/download/v1.1.1/terragrunt_linux_amd64
           chmod +x /usr/local/bin/terragrunt
 
       - uses: actions/setup-go@v5
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -12,6 +12,11 @@ self-contained deploy kit and its runbook in [`azure-config/README.md`](./azure-
 # OpenTofu drives both layers locally — the physical layer requires >= 1.11.1, and
 # OpenTofu runs the logical layer too. Terragrunt orchestrates them.
 brew install opentofu terragrunt
+# Terragrunt matters here: the harness and these docs use the post-0.78 CLI
+# (`run --all`, `--non-interactive`). The old `--terragrunt-*` flags were REMOVED in
+# 0.85.0 with no grace period, so an old binary fails with "flag provided but not
+# defined". Check with `terragrunt --version`; if brew gives you something older,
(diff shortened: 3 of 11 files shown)


ddvdozuki:
## Resolved: Terragrunt 1.1.1 drives Terraform 1.5.7 fine

The open question blocking the infra-live bump was whether `terragrunt_version = "1.1.1"` is safe for the Vault stacks, since that setting applies to *every* stack and the Vault ones run `TERRAFORM_FOSS` 1.5.7 (the MPL holdout).

Tested with the real binaries — terragrunt 1.1.1 pointed at Terraform 1.5.7:

```
$ TG_TF_PATH=.../1.5.7/terraform terragrunt-1.1.1 init --non-interactive
Terraform has been successfully initialized!          exit 0

$ ... terragrunt-1.1.1 plan --non-interactive
Plan: 1 to add, 0 to change, 0 to destroy.            exit 0

$ ... terragrunt-1.1.1 run -- version
Terraform v1.5.7
```

Terragrunt detects the binary correctly (it labels output `terraform:` rather than `tofu:`) and both init and plan work. So the Vault stacks can take the terragrunt bump without leaving 1.5.7, and `stacks.tf` needs no special-casing beyond the `is_vault` split it already has.

Remaining gate on the infra-live change is now just the harness validation cycle on tofu 1.12.5 + terragrunt 1.1.1.

ddvdozuki:
## OpenTofu 1.12.5 pre-check: no new breakage in either layer

Before this becomes an infra-live change, checked that 1.12.5 doesn't upset the modules. Bare-module `validate` always reports artifacts (terragrunt generates the `aws.dr` provider block and the vault `address`), so the count only means something compared against the version we run today:

| module | tofu 1.12.2 (current local) | tofu 1.12.5 (target) |
|---|---|---|
| `terraform/physical` | 29 errors | **29 errors** |
| `terraform/logical` | 1 error | **1 error** |

Identical, and every error is a pre-existing artifact of validating the module without terragrunt's generated providers — not a regression.

This is a static check, not a plan. The real test is a full harness cycle on tofu 1.12.5 + terragrunt 1.1.1, which is still the gate on the infra-live bump. Running it requires switching the active binaries, and `tgenv`/brew switch globally, so it waits for the in-flight DR validation cycle to finish.

PR #376, opened 2026-07-30
feat(bi): BI database on Aurora Serverless v2, and default the RDS engine to 8.4

ddvdozuki:
Gets new capacity off MySQL 8.0, which left standard support on **2026-07-31** and now bills RDS Extended Support per vCPU-hour.

## The BI database was the real exposure

`bi.tf` pinned the DMS-target BI database to `engine_version = "8.0"` with **no input to change it**, so every stack got an 8.0 BI database even when its primary was Aurora 8.4. Unlike the RDS primaries, nothing retires these. Across the 3M fleet that is six instances, ~42 billed vCPU, on the order of $3k/month indefinitely.

`bi_db_engine` selects the engine. `"aurora"` provisions an Aurora MySQL Serverless v2 cluster on `var.aurora_engine_version` (8.4, supported into 2029), scaling to `bi_aurora_min_acu` (0.5) between queries. BI is bursty and idle most of the time, where a provisioned instance bills flat around the clock, so this cuts steady-state cost as well as the Extended Support charge. The code default stays `"rds"`: stacks that predate the variable keep what they have, and new stacks get aurora from `_templates/env.hcl` (infra-live #165).

Consumers now read engine-agnostic locals (`bi_db_host/port/username/password/identifier/resource_id`), so the DMS target endpoint and the BI credentials secret follow whichever engine is active. Replaces the dead `local.bi_db`, which was defined and never referenced.

Existing stacks are held on the rds path by the default. Switching engines replaces the database and db-replace-guard blocks that without the label. The BI database is a rebuildable DMS target (a full load repopulates it), so the flip is safe to schedule; cheapest alongside that env's primary KMS key swap, since both replace the same instance.

Separate from the engine flip: the first apply that picks up this release replaces the provisioned replication instance + task with the serverless config on **every** DMS-enabled stack, engine pinned or not. No manual pre-step is needed; the locked provider (v6.56.0) stops the old task and waits before deleting it. The replication is created with `start_replication = false`; the `dms-start` Job in the logical layer starts it after migrations, and the first start is a fresh full load (drop-and-create), so BI has a reload window at that bump. `util/dms-stop.sh` stays in the tree as an optional pre-stop for anyone who wants the old task fully stopped before the new load begins.

## RDS engine default 8.0 -> 8.4

Default only. Nothing in infra-live pins `rds_engine_family`/`rds_engine_version`, and that is a decision rather than a gap: the remaining 8.0 instances are either Aurora-migration rollback anchors headed for deletion, or the last two live primaries (usac, qa) whose route to 8.4 is their imminent Aurora migration. The default only bites a stack that bumps `infra_version` while still on provisioned 8.0, so those two envs migrate before they bump.

Verified inert beyond the version: `db.m5.large` is orderable on 8.4.x in us-east-1 exactly as on 8.0.43, so `data.aws_rds_orderable_db_instance` resolves the same class and the `rds_instance_memory` postcondition still holds. The lookup itself needed a review fix: DescribeOrderableDBInstanceOptions rejects a bare family ("Engine version is not a valid full version"), and the data source has no count, so the `"8.4"` default failed every plan, aurora stacks included. `aws_rds_engine_version` with `latest = true` now resolves the family to a concrete minor first; a full pin passes through unchanged.

## One trap worth recording

Both legacy modules use `var.rds_engine_family` (the bare `"8.0"`/`"8.4"` **prefix**), not an exact minor. AWS resolves a prefix to the latest matching minor, which is what lets `auto_minor_version_upgrade` move these without Terraform fighting it. An exact pin would have been actively dangerous here: the old default was `8.0.43` while live instances have drifted to `8.0.46`, so it would plan a **downgrade** that AWS rejects, failing every apply. I made that mistake mid-change and caught it against the live inventory.

## Verification

- `tofu fmt`, `terraform validate with tflint`, `terraform-docs` all pass (the repo's own pre-commit set).
- `tofu validate` passes with the two Terragrunt-generated provider aliases stubbed.
- Live inventory cross-checked for every affected instance and version.
- Provider v6.56.0 source checked: `aws_dms_replication_task` delete stops a running task and waits for stopped, so the task-to-serverless swap applies cleanly with no manual stop.

## Review rounds 4 and 5

Eight findings across the first three rounds were already folded in above. Rounds 4 and 5 found seven more; all are fixed on the branch.

- **Fresh-install ordering was a permanent wedge, not a quirk.** The replication originally started at the physical apply with `start_replication = true`, but its selection rules match schemas the logical layer's migrations create, and physical runs first. On a fresh stack DMS refuses to initialize ("No tables were found"), the provider waits on the start and fails the physical apply, and logical (which would create the schemas) can never run. Upgrade-path stacks never hit it because their schemas exist, which is why every rehearsal passed. Fixed by `start_replication = false` plus bringing back the `dms-start` Job in the logical layer, taught to dispatch on the ARN shape (serverless config vs legacy task). Its deployment wait always was this ordering gate.
- Consequence to keep in mind: the provider's update path stops a serverless replication before modifying it and only restarts it when `start_replication` is true. A physical modify therefore leaves BI stopped until the next logical apply starts it. Logical autodeploys after physical, so this closes on its own, but a physical apply must never be left un-followed: a serverless replication stopped for 48 hours is deprovisioned and can only be recreated.
- The `dms-start` Job is generation-named: physical hashes the endpoint ARNs, mappings, settings, and DCU bounds into `dms_replication_generation`, and logical folds that into the Job name, so any modify that stopped the replication gets a fresh Job to restart it. A completed Job with a static name would have sat inert.
- The migration runner's BI probes failed open in two layers: a denied or throttled describe call classified a serverless stack as a legacy task (and the cached result poisoned the whole run), and a `die` inside command substitution only killed the subshell. Both fixed: probes fail closed, and kind/ARN resolve once in the parent shell before phase dispatch.
- The Aurora BI cluster now carries the same log exports, retained log groups, and audit/slow/general parameters as the primary Aurora path; switching BI engines no longer drops the 365-day trail.
- The upgrade harness treats a known-but-never-started replication config as "created" and polls, instead of hard-failing before the Job's first start.
- The engine-switch runbook gained the out-of-band `--no-deletion-protection` pre-step; Terraform cannot clear it in the same apply that deletes the instance.

## Still to observe on a qa canary before fleet rollout

- A genuine fresh install starting the replication from the `dms-start` Job after migrations, in the right order, with the physical apply staying green.
- The EventBridge rule matching serverless events on the replication-config ARN in `resources`, and the restart lambda's "has failed" wording against a real event.
- The Slack link's `#serverlessReplicationDetails/` console route resolving.

## Review round 6

A final verification pass (independent agent plus codex) confirmed all rounds 1-5 fixes present, then found six more items; all fixed in `9ef5ef9..08b3dcc`.

- The engine-switch docs claimed the rds->aurora repoint mints a fresh `dms-start` Job. It cannot: the target endpoint is modified in place, its ARN is stable, and the replication config sees no diff, so the generation hash never moves. The operator owns the restart on any endpoint repoint (and it must be `reload-target`, which is also why endpoint attributes stay out of the hash). Docs corrected.
- The generation hash now covers the whole `compute_config` (subnet group, security groups, kms key, multi_az), not just the DCU bounds - the provider stops the replication on any of them, so each needs to mint a restart Job.
- The `dms-start` Job no longer exits clean on `modifying`/`stopping` (it polls up to 10 minutes and acts on the resolved state) and treats `deprovisioning` as fatal.
- A deprovision event now pages the channel; it was posting as routine noise, and it is the one unrecoverable state.
- The `bi_ha` CI config now exercises the aurora BI path end to end (it was previously building no BI database at all); `full` keeps the provisioned rds path covered.
- `dms-stop.sh` learned the serverless API, the runner's BI stop waits are bounded and fail on `failed`, and the harness allows the serverless start path its real ~40 minute greenfield worst case.


ddvdozuki:
### Code review

Found 5 issues:

1. The RDS engine default bump applies an unplanned major-version upgrade to live primaries. `rds_engine_version`/`rds_engine_family` move 8.0 -> 8.4, but no infra-live stack pins either variable (checked all six `db_engine = "rds"` stacks: gca, apac, latam, emea, qa, usac). `rds.tf:3` feeds this straight into the primary `aws_db_instance` and `rds.tf:173` sets `allow_major_version_upgrade = true`, so the next routine `infra_version` bump *performs* an in-place 8.0 -> 8.4 upgrade rather than failing. `db-replace-guard` does not catch it: this is a `ModifyDBInstance`, not a replace. The variable description's claim that existing stacks pin "8.0" does not hold today.

https://github.com/Dozuki/CloudPrem-Infra/blob/0eb02ddf503f9c679ae85702a9c501eafa7d4bf2/terraform/physical/variables.tf#L246-L252

2. The `dms_task_arn` output changes shape without updating its consumer (terraform/physical/CLAUDE.md says "Changing any output? Update `../CONTRACT.md` and the logical-layer consumer together."). It now returns a `replication-config:` ARN instead of a `task:` ARN, unconditionally for every `dms_enabled` stack. `terraform/logical/bi.tf`'s `kubernetes_job_v1.dms_start` still calls `aws dms describe-replication-tasks --filters Name=replication-task-arn,Values=$ARN`, which matches nothing and hits its own `None|"") ... exit 1`. Because that Job sets `wait_for_completion = false`, the apply reports green while DMS silently never starts. `terraform/CONTRACT.md` is also untouched.

https://github.com/Dozuki/CloudPrem-Infra/blob/0eb02ddf503f9c679ae85702a9c501eafa7d4bf2/terraform/physical/outputs.tf#L43-L48

3. The restart lambda lacks IAM permission for the API it now calls. `dms_restart.py` calls `dms:StartReplication` for serverless ARNs, but `lambda_permissions` grants only `dms:DescribeReplicationTasks` and `dms:StartReplicationTask`. Every serverless auto-restart will fail AccessDenied, which disables the recovery path the new comments advertise. The `dms-start` Job's Pod Identity policy in `eks.tf` has the same task-only scope.

https://github.com/Dozuki/CloudPrem-Infra/blob/0eb02ddf503f9c679ae85702a9c501eafa7d4bf2/terraform/physical/monitoring.tf#L19-L24

4. `start_replication = true` preempts the documented `reload-target` cutover, leaving the new BI database without its historical data. On the rds -> aurora switch the apply changes `target_endpoint_arn`; the provider's `resourceReplicationConfigUpdate` then calls `startReplication()`, which selects `resume-processing` whenever status is not `ready` (i.e. the stopped state the runbook in `bi_db_engine` tells the operator to leave it in). CDC resumes from the last position against an empty Aurora target, so the full load never runs and the operator's manual `reload-target` step is bypassed rather than being what starts it.

https://github.com/Dozuki/CloudPrem-Infra/blob/0eb02ddf503f9c679ae85702a9c501eafa7d4bf2/terraform/physical/bi.tf#L161-L165

5. The change silently breaks `aurora-migration-runner.sh`'s BI fence. `bi_task_status()` resolves the BI task via `aws dms describe-replication-tasks`, which no longer returns anything once the BI replication is a `replication_config`. The `fence` phase's `if [ "$(bi_task_status)" = "running" ]` guard then never fires, so the BI replication is no longer stopped before the cutover apply repoints its source endpoint - the race that guard was added to close. `abort` and `bi-epoch` break the same way. emea is mid-cutover and usac has not started.

https://github.com/Dozuki/CloudPrem-Infra/blob/0eb02ddf503f9c679ae85702a9c501eafa7d4bf2/terraform/physical/bi.tf#L140-L166

PR #383, opened 2026-07-30
test(harness): run the harness on Argo - image, publish path, workflow templates

ddvdozuki:
Phase 2 of moving the harness onto Argo Workflows. The harness only runs from a laptop today; this is the runtime that lets a phase run as a pod. `run.sh` stays the local launcher and drives the identical subcommands, so the two paths stay comparable.

Nothing here changes harness behavior. All four files are new and wholly inside `live/tests`, so release-please still excludes them from the product version.

## The point of the build is deleting `tofu init`

The harness runs terragrunt once per layer per git worktree, so an upgrade scenario pays `init` **four or more times**, each resolving ~12 providers. Nothing in the repo configures a provider cache today, so every one of those inits downloads the whole set again. Stage 3 warms a plugin cache from both layers' committed lockfiles with `tofu init -backend=false`, which needs no AWS credentials and so works in CI.

Measured in-container, logical layer, clean working dir, linux/amd64:

| | wall time | disk written |
|---|---|---|
| `tofu init` **with** the baked cache | **6.6s** | **148KB** (11 symlinks into the cache) |
| `tofu init` without | 85s | 1.3GB |

Every provider symlinks into `/opt/tofu-plugin-cache`, so the cache costs disk once in the image rather than once per worktree. At 4 inits per upgrade run that is roughly **5 minutes and 5GB of writes saved per run**, and the cold number came off a fast home connection, so a pod egressing through NAT to the public registry will do better than that.

The harness builds terragrunt's env with `append(os.Environ(), ...)`, so `TF_PLUGIN_CACHE_DIR` as an image `ENV` was all it took. No Go change.

Best-effort on purpose: the cache is keyed by provider version and the harness also tests old refs, whose lockfiles may pin versions it lacks. A miss falls back to a normal download, so a stale cache costs time, never correctness.

## What the entrypoint does

The server-side half of `run.sh`, minus the laptop ergonomics (SSO runway checks, the Vault port-forward, the Little Snitch workaround):

- arms the az-shim (baked at `/opt/az-shim`, so the azurerm-on-AWS class of bug still surfaces in a run); `NO_AZ_SHIM=1` drops it
- clones or **fetches** the repo. Mount `HARNESS_REPO_DIR` as a run-scoped volume and every phase after the first is an incremental fetch. Needs a real clone with tags and `origin/*` refs, because the harness builds a worktree per ref
- mints the manual-TLS cert (`TF_VAR_tls_cert`/`_key`)
- logs into Vault with the pod's own identity. No port-forward and no second SSO profile: the Argo cluster reaches `vault.internal.dozuki.com` over the PrivateLink endpoint in its own VPC. `VAULT_AWS_REGION` is there for cross-partition STS on gov
- prints the toolchain, then execs the phase

Anything that is not a phase name is exec'd as given, so `bash` and `cpi-dr` both work.

## Verified

- builds clean; `uname -m` = `x86_64` and the ELF header confirms amd64 (`docker inspect .Architecture` reports the host arch and is not proof)
- all 12 tools resolve; tofu 1.11.8 + terragrunt 0.99.4 match the fleet pin in `_spacelift/stacks.tf`
- az-shim exits 1
- full entrypoint path end-to-end with no AWS: clone -> detach at master -> TLS -> banner -> `harness teardown` resolving the default `--matrix` relative to the repo and no-op'ing cleanly on a missing manifest
- every tool download checksum-verified against the publisher's SHA256SUMS. This image carries admin credentials for a whole account

## Size, and the next 700MB

Rootfs is 2.4GB: 1.3GB provider cache (the trade above), 537MB **`vault`**, ~180MB python+`aws-cli`, 315MB tofu/terragrunt/helm/kubectl, 96MB harness binaries.

The `vault` CLI is 537MB for exactly two operations, `vault login -method=aws` and `vault token renew`, because the binary ships the whole server and UI and there is no CLI-only build. It is larger than every other tool combined. Dropping it and the `aws` CLI (4 call sites) for API calls the harness already has libraries for would land the image near **1.7GB**. Both are follow-ups, not done here, and `IMAGE.md` records the call sites.

Alpine is the base specifically because its community `aws-cli` is musl-native; the official v2 installer is glibc-only, which would force Debian and ~230MB more.

## Not wired yet

The registry. `IMAGE.md` recommends the **dozukicloud** ECR (069) with cross-account pull via the existing `dozukicloud-ecr-policies` stack, rather than an ECR repo in the management account: same-account pull would be marginally faster, but it means trusting GitHub OIDC inside the account that holds Vault, and the existing precedent keeps GitHub-trusted roles in sandbox/platform accounts. Still to wire: the repo, a pull-policy entry for the Argo pod identity, and a build-and-push workflow pinning by digest.


ddvdozuki:
Extended this PR with the rest of the runnable path, since the pieces only make sense together (the templates reference the image, and the image gained the cross-account identity the templates rely on).

## Added

**`live/tests/argo/`** - the WorkflowTemplates.

- `harness-phase`: one template for all four subcommands. `retryPolicy: OnError`, deliberately **not** `OnFailure` - OnError covers the pod dying (eviction, OOM, spot reclaim, node replacement), which re-entrancy makes safe to repeat, while a non-zero exit from the harness is a real verdict and retrying it would burn another hour to report the same thing.
- `harness-scenario`: one config end to end, **teardown as an exit handler** rather than a final step, because a final step does not run when an earlier step fails, which is exactly when teardown matters. And because teardown reads the run manifest from S3, cleanup is no longer tied to the process that created the stack. That is the structural fix for `run.sh`'s trap, which failed on the 2026-07-29 interrupt and needed a separate manual sweep.
- `harness-matrix`: fans out and submits one child **Workflow** per config rather than looping inside one workflow. Costs a resource template and a small RBAC role (the chart's workflow RBAC cannot create Workflows), buys per-config isolation: own UI, own retry budget, own resume scope, own exit handler. "Re-run just `full`" is resubmitting one child.
- **One semaphore across every entry point.** DDVtest allows 10 VPCs per region and a stack is about one (recovery is two), so a nightly matrix plus a couple of PR runs would exhaust the quota mid-apply and surface as a confusing error deep inside a terragrunt run. Children hold it provision-through-teardown; the orchestrator deliberately does not take it, because gating both would deadlock any matrix larger than the semaphore.

**Entrypoint: cross-account identity.** The pod's role is in the management account, the test stack is in DDVtest. Rather than teaching the harness to chain roles, it writes one named profile whose `credential_source` is the Pod Identity endpoint EKS already injects, so every tool assumes the DDVtest role with no code change - including the harness's existing `aws configure export-credentials --profile` path, which exists for exactly this. The Vault login moves **ahead** of it: Vault AWS-auth must see the pod's own identity, not the assumed DDVtest role, and ordering keeps them apart the way `run.sh` uses a subshell.

**`.github/workflows/harness-image.yml`** - build and publish to the ECR repo from infra-tf#69 / infra-live#169, via a GitHub-OIDC role scoped to `refs/heads/master`. PRs build without publishing so a Dockerfile change is proven before merge.

## Two prerequisites, deliberately not done here

The templates **cannot run** until both exist, and both need the Argo workflow role ARN, which only exists after the `ArgoOpsCluster` stack applies:

1. The DDVtest `cloudprem-upgrade-tests` role's trust policy must admit the Argo workflow role. Extending the existing role beats minting a second one - it keeps one identity for "the thing that drives DDVtest".
2. That role must be registered as a Vault AWS-auth role, in infra-tf `vault-config`, which is applied by hand over a port-forward.

`live/tests/argo/README.md` states both, plus that the **recovery scenario stays on `run.sh`**: `cmd/harness` has no phase for `RunRecovery`.

## Verification

All three manifests parse, including the embedded child Workflow manifest inside the resource template. `actionlint` clean on the workflow, `shellcheck` clean on the entrypoint. Nothing here has been applied to a cluster - there is no cluster yet.

Commits are split so `live/tests` changes stay wholly inside `live/tests` (release-please `exclude-paths`) and the `.github/` change is a separate `ci:` commit.

PR #404, opened 2026-08-01
fix(monitoring): stop alerting on routine DMS serverless scaling events

ddvdozuki:
## Problem

A DMS serverless replication emits an EventBridge event for every autoscaling
decision it makes, including the ones where it decides it *cannot* act:

> The replication, '<name>', cannot scale down as the replication is already at the
> provided Minimum DMS Capacity Units, '2'.

On an env in steady state these are the only DMS events there are, so this is the
whole of what the alert channel sees from DMS. It trains everyone to scroll past
DMS posts, which is a problem the day one of them is a failure.

## Changes

**`util/sns_to_slack.py`** - drop the routine serverless lifecycle messages: the
scaling decisions (up, down, completed, blocked at min, blocked at max) and the
provisioning-pipeline steps (initializing, fetching metadata, calculating capacity,
provisioning, provisioned, modifying).

Still posts: failed, stopped, being stopped, started, running, created, deleted,
deprovisioning, deprovisioned. Aurora-migration `ReplicationTask` events, which share
this rule and this lambda, are untouched by the list.

Two things worth reviewing:

- The drop is evaluated *after* the failure tokens, not before. `"provisioning its
  capacity"` is a substring of `"deprovisioning its capacity"`, and `"has been
  provisioned"` of `"has been deprovisioned"` - deprovisioning is the one BI state
  nothing recovers from, so it must not be swallowed by its own prefix. Ordering is
  what prevents that.
- It is a denylist, not an allowlist: a message AWS words differently than we assumed
  still posts. Fail-open, same reasoning as the EventBridge rule not pinning
  `detail-type`.

It also runs before `get_task_name`, so suppressed events no longer spend a
`DescribeReplications` call each.

**`monitoring.tf`** - new `<identifier>-dms-capacity-saturated` alarm. Dropping the
per-event scaling posts loses the one thing they gestured at, real capacity pressure,
and no single event can tell you it is sustained. `CapacityUtilization >= 90%` for
12x300s does. AWS ties this metric to the failure mode directly: a DCU floor or
ceiling set too low for the workload shows as utilization "consistently at its maximum
value" and can fail the replication with an OOM. So the alarm names a config change
(raise `bi_dms_min_dcu` / `bi_dms_max_dcu`), not a transient.

The hour is deliberate. A full load legitimately runs hot and scale-up is not instant,
so a shorter window pages on healthy bursts - the exact noise being removed.
`treat_missing_data` is `notBreaching` here because the two CDC latency alarms are
already `breaching` for "not reporting"; a third would just double-page every
deprovision and every config replacement.

## Testing

29 classification cases over the real AWS message strings for both the `Replication`
and `ReplicationTask` source types (docs `CHAP_EventBridge.html`), covering every
drop, both substring traps, and the task events that must be unaffected. All pass.

`tofu fmt` clean. `tofu validate` unchanged from baseline (29 pre-existing errors both
before and after - the `dr` provider alias is caller-injected, so a standalone module
validate cannot resolve it).

## Blast radius

Alerting only. No replication resource is touched, so this carries no stop/start of
any replication.

### DIFF
--- a/terraform/physical/README.md
+++ b/terraform/physical/README.md
@@ -66,6 +66,7 @@
 | [aws_cloudwatch_metric_alarm.aurora_memory](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_metric_alarm) | resource |
 | [aws_cloudwatch_metric_alarm.bi_cdc_latency_source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_metric_alarm) | resource |
 | [aws_cloudwatch_metric_alarm.bi_cdc_latency_target](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_metric_alarm) | resource |
+| [aws_cloudwatch_metric_alarm.bi_dms_capacity_saturated](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_metric_alarm) | resource |
 | [aws_cloudwatch_metric_alarm.dr_replication_deadman](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_metric_alarm) | resource |
 | [aws_cloudwatch_metric_alarm.dr_s3_replication_failed](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_metric_alarm) | resource |
 | [aws_cloudwatch_metric_alarm.dr_s3_replication_latency](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_metric_alarm) | resource |
--- a/terraform/physical/monitoring.tf
+++ b/terraform/physical/monitoring.tf
@@ -518,6 +518,41 @@ resource "aws_cloudwatch_metric_alarm" "bi_cdc_latency_target" {
   tags                = local.tags
 }
 
+# Capacity saturation. This is the alarm that lets the slack lambda drop the per-event
+# scaling chatter: a serverless replication posts every scale up, scale down and
+# scale-blocked decision, and no single one of those tells you whether the replication is
+# actually short of capacity. An hour of CapacityUtilization pinned at 90%+ does.
+#
+# AWS's own guidance ties this metric to the failure mode: a min or max DCU set too low
+# for the workload shows up as CapacityUtilization "consistently at its maximum value",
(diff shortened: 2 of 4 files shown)


ddvdozuki:
Ran a multi-model review over this and infra-live#178. Four findings landed on this PR; all are now fixed in 5f09736.

**`cannot scale up` should never have been on the denylist.** It reads like its scale-down twin and is the opposite: at-minimum means we are paying for capacity nobody wants, at-maximum means the ceiling is throttling the replication. No failure token, so listing it silently discarded the strongest point-in-time evidence that `max_capacity_units` is binding - exactly what the new alarm otherwise needs an hour of datapoints to infer. With max at 32 against a 1-2 DCU steady state it should about never fire, so dropping it bought no quiet at all. Removed.

**`notBreaching` -> `ignore` on the capacity alarm.** `notBreaching` was right that the CDC latency alarms already own "stopped reporting", but wrong the other way: with this alarm already in ALARM, a replication that dies and stops emitting resolves it ALARM -> OK and posts a recovery for something that is dead. Confirmed against the CloudWatch missing-data docs - `ignore` maintains current state.

**The alarm description claimed more than the alarm can show.** `Maximum` with 12/12 datapoints means a sample touched 90% in each of twelve 5-minute buckets. That is repeated peaking, not an hour above 90% - a replication oscillating 40%/91% trips it. Reworded. Also documented the full-load blind spot: on the one full load observed (gca 07-31 19:18-19:24Z) DMS published no `CapacityUtilization` at all until 20:45Z, so the alarm may be silent exactly when utilization peaks. Scoped to that observation rather than asserted as general DMS behaviour, and noted that the gap is covered by the failure event rather than here.

**The fixture matrix is now committed** as `util/test_sns_to_slack.py` - 38 cases across both source types, no deps, `python3 util/test_sns_to_slack.py`. This is the durable one. The routing is only correct because the failure tokens are evaluated before the denylist, which is what stops `"deprovisioning its capacity"` being eaten by the `"provisioning its capacity"` entry that prefixes it. Nothing at the call site shows that, and a tidy-up would break it silently. The lambda `archive_file` names a single source file, so the test does not ship in the zip.

Also tightened `"is being tested"` to `"connections tied to"`, which was the loosest token in the list.

`tofu fmt` clean; `tofu validate` unchanged from baseline (29 pre-existing errors before and after - the `dr` provider alias is caller-injected).

Still draft: this needs to merge and cut a release before infra-live#178 can go, since that PR's environments are pinned to `v8.7.0` and this is not in it. See the note on #178.

PR #424, opened 2026-08-02
feat(logical): harden external-secrets webhook, pin upgrade/rollback force=false

ddvdozuki:
Two independent commits, both fallout from the dev-min incident on 2026-08-01.

## 1. external-secrets webhook HA

The webhook runs 1 replica with no PDB, no spread and no resource requests (all chart defaults). Its ValidatingWebhookConfiguration is fail-closed, so a single pod reschedule fails admission for every ExternalSecret and SecretStore in the namespace. On dev-min that took down 6 ExternalSecrets plus the SecretStore at once and made the rollback loop materially worse.

- `replicaCount: 2`
- PDB with `maxUnavailable: 1`. Not `minAvailable: 2`, which on a 2-replica deployment forbids every eviction and permanently blocks Karpenter consolidation of any node hosting a replica. At 2 replicas `maxUnavailable: 1` and `minAvailable: 1` are arithmetically identical anyway.
- Soft (`ScheduleAnyway`) topologySpreadConstraint by hostname. Not `DoNotSchedule`, which would strand the 2nd replica Pending forever on genuinely single-node clusters. Soft means preference, not guarantee: on a one-node cluster both replicas still land together. `topologyKey` is hostname, so this is node-loss protection only and buys nothing against a zone failure.
- Requests `10m` / `64Mi`. Measured live on dev-min at 1m / 32Mi; memory is requested above observed to leave headroom.
- `priorityClassName: system-cluster-critical`.

No limits, on purpose. An under-set memory limit OOMKills the webhook, which is the outage this is meant to prevent.

### What the requests and the priority class actually do

An earlier revision of this PR (and of the code comments) claimed the requests are what make the topology spread meaningful, and that the priority class lets an evicted replica preempt instead of waiting. Both overstated the mechanism. Corrected:

**Requests** do not affect the spread at all. `PodTopologySpread` scoring counts matching pods per topology domain and never looks at resource requests. What real requests buy is separate and still worth having: a zero-request pod lands in BestEffort QoS, which is the first thing the kubelet evicts under node pressure, and Karpenter treats it as free in bin-packing, so it can be packed onto a node with no real headroom left.

**The priority class** shortens the Pending window but does not remove it. A Pending replica keeps the PDB blocked, since `disruptionsAllowed` counts healthy pods and not scheduled ones, so one unschedulable replica stalls every drain behind it. Preemption only fires when evicting lower-priority pods on some node actually makes this pod fit, and at a 10m/64Mi request it usually fits somewhere without preempting anything. On a genuinely full cluster it still waits on Karpenter. What the class buys unconditionally is the rest: head of the scheduling queue, nothing else can preempt it, and the kubelet ranks it last for node-pressure eviction. That is a fair trade for a fail-closed admission webhook, where every ExternalSecret and SecretStore in the cluster is stuck behind it.

## 2. `upgrade.force` and `rollback.force` pinned false on the dozuki HelmRelease

Both are Helm's default. They are written down anyway because the semantics are widely misread, including by two earlier revisions of this PR.

**What `force` actually does.** Checked against source, not docs: helm v3.19.0 `pkg/kube/client.go` `updateResource` and v4.0.0 `replaceResource` both take the same branch, `helper.Replace`, which is a whole-object PUT. It does **not** delete and recreate. Earlier revisions of this description and of the code comments claimed a completed `db-migrations` Job would be deleted and re-run against a database one schema version ahead. That is wrong, and the corrected reason to keep force off is close to the opposite:

A PUT sends only what the chart renders. Any field the server populated and the chart does not carry is dropped or makes the write fail outright, and an immutable field whose live value was server-defaulted turns a survivable patch into a hard failure. The auto-stamped `batch.kubernetes.io/controller-uid` Job selector is exactly that case, which is why reaching for `force` to punch through an immutable-field error makes it worse rather than better. It is an all-or-nothing write with no upside on this release.

Neither pin governs the ordinary create/prune cycle that happens when the Job name changes.

**Scope, honestly stated.** helm v4 refuses force-replace together with server-side apply, and the HelmRelease CRD documents `force` as ignored under SSA. The `flux_chart_version` pinned here is on the SSA line, so today both pins are defense-in-depth for client-side or adopted releases rather than an active guard.

`upgrade.force` is the one that was missing, and it is the more exposed of the two: every reconcile that produces a change goes through upgrade, while rollback only runs on remediation.

## Verification

- Rendered ESO 2.8.0 with the exact `set` list. Webhook Deployment comes out with `replicas: 2`, the requests, `priorityClassName: system-cluster-critical`, and the topologySpreadConstraint with `labelSelector` auto-filled from the chart. PDB renders `maxUnavailable: 1`. All five keys confirmed present under `webhook:` in the chart's values.
- `kubectl explain helmrelease.spec.upgrade.force` against live dev-min confirms the field exists on the `helm.toolkit.fluxcd.io/v2` HelmRelease and confirms the SSA caveat above.
- `tofu fmt -check -recursive` clean.

## Blast radius

Logical layer, so every AWS/Azure logical stack on the next `infra_version` bump. Airgapped onprem installs are unaffected: they don't get external-secrets from this layer.

## Ordering

The `force` pins are independent of the companion chart change and can land first. Nothing here depends on that chart change, and nothing in it depends on this.

### DIFF
--- a/terraform/logical/flux.tf
+++ b/terraform/logical/flux.tf
@@ -547,9 +547,16 @@ resource "kubectl_manifest" "dozuki_helmrelease" {
       # upgrade retries=2: with retries=0 ONE failed upgrade (an eviction, a
       # not-yet-mirrored image, a transient webhook) stalls the release until a
       # human does the requestedAt+forceAt dance (seen live 2026-07-30). Two
-      # retries let it self-heal once the cause clears; remediateLastFailure
-      # stays false so a failure never auto-rolls-back over a completed
-      # migration. Install keeps retries=0: install remediation uninstalls
+      # retries let it self-heal once the cause clears. What that costs:
+      # remediation (strategy defaults to rollback) runs BETWEEN each attempt,
+      # so retries=2 means up to 2 rollbacks per failed upgrade, and the counter
+      # resets on a chart version change, so it bounds one cycle and not the
+      # loop (dev-min 2026-08-01). remediateLastFailure=false is load-bearing,
+      # not redundant: Flux defaults it to TRUE whenever retries > 0. It governs
+      # only the failure after retries are spent, parking the release on the new
+      # revision instead of rolling back - a human sees a failed release rather
+      # than a chart silently older than the schema already in the database.
+      # Install keeps retries=0: install remediation uninstalls
       # first, which is not a loop we want running unattended.
       upgrade = { disableWait = false, timeout = "4h30m", crds = "CreateReplace", remediation = { retries = 2, remediateLastFailure = false } }
       # rollback.force=false pinned explicitly (it is the Helm default, but this is the one
--- a/terraform/logical/flux.tf
+++ b/terraform/logical/flux.tf
@@ -553,20 +553,23 @@ resource "kubectl_manifest" "dozuki_helmrelease" {
       # resets on a chart version change, so it bounds one cycle and not the
       # loop (dev-min 2026-08-01). remediateLastFailure=false is load-bearing,
       # not redundant: Flux defaults it to TRUE whenever retries > 0. It governs
-      # only the failure after retries are spent, parking the release on the new
-      # revision instead of rolling back - a human sees a failed release rather
-      # than a chart silently older than the schema already in the database.
+      # only the failure after retries are spent: Flux performs no rollback, so the
+      # release stays failed and live state may be left partially applied, pending
+      # operator recovery. A human sees a failed release rather than a chart
+      # silently older than the schema already in the database.
       # Install keeps retries=0: install remediation uninstalls
       # first, which is not a loop we want running unattended.
       upgrade = { disableWait = false, timeout = "4h30m", crds = "CreateReplace", remediation = { retries = 2, remediateLastFailure = false } }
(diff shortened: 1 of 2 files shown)


ddvdozuki:
Added a third commit (`fa2769e`), comment-only, no behaviour change.

The existing comment on the upgrade block claimed `remediateLastFailure: false` means "a failure never auto-rolls-back over a completed migration". It doesn't. Per the [helm-controller v2 API](https://fluxcd.io/flux/components/helm/api/v2/), remediation is "performed between each attempt" and `strategy` defaults to `rollback`, so `retries: 2` permits up to **two** rollbacks per failed upgrade. The flag only governs the failure after retries are spent. That is exactly the ping-pong on dev-min 2026-08-01, and the failure counter resetting on a chart version change is why `retries: 2` didn't cap it at two.

Worth flagging what the docs also say: *"Defaults to `false` unless `Retries` is greater than 0."* Since we run `retries: 2`, Flux's own default here is **true**. The explicit `false` is load-bearing rather than redundant, and dropping it would add a third rollback.

Value is unchanged and stays `false`. With the migration Job selector fixed in Dozuki/helm#213, rollback can actually succeed, so `retries: 2` behaves as originally intended.

PR #426, opened 2026-08-02
feat(physical): shrink and harden the aurora migration cutover window

ddvdozuki:
Three related changes to the aurora migration cutover path. All measured, none wired into the fence's correctness logic without evidence.

## 1. Fence pre-flight gate on backups and in-flight applies

`fence` refuses to start when a parameter apply would be deferred: instance not `available`, an apply already in flight, a snapshot creating, or the backup window open / within `FENCE_BACKUP_MARGIN_MIN` (default 30) minutes.

The `read_only` fence normally lands in **3.3-6.7s** (measured in ddvtest on RDS 8.0.46, both db.t3.medium and db.m5.8xlarge). One run took **262.89s** because a backup was running: RDS queues the parameter apply behind the backup, then applied it 9s after it finished.

The duration is not the real problem. For that whole window `ModifyDBParameterGroup` refuses the unfence with `InvalidDBParameterGroupState` (reproduced deterministically), so a **backup-contended fence is an unabortable fence**: source read-only, customers down, documented rollback unavailable until the backup completes.

Checked twice, on entry to fail fast and again immediately before the apply prompt, since the BI stop, kill sweep and marker commit sit between them. `status` runs the same check read-only so a window can be picked in advance. `abort` now reports the parameter group state before prompting for the unfence, so a refused unfence reads as the expected self-clearing condition rather than a mystery Terraform failure.

Deliberately does **not** touch `BackupRetentionPeriod`. Setting it to 0 to suppress automated backups also deletes the existing ones, which are the rollback path for the cutover being run. `FENCE_PREFLIGHT=warn` overrides so a stuck or misreported backup can never strand a cutover that has to proceed.

## 2. DMS status control table + `proof-observe`

The drain proof has to STOP the DMS task, because `RecoveryCheckpoint` is a stopped-task safepoint: it advances to the true applied position only on stop, not while running. That stop plus up to 10 resume/stop drain cycles sits inside the write freeze.

`awsdms_status` is the one target-side progress view that needs no stop. It carries `SOURCE_CURRENT_POSITION` (how far DMS has read) and `SOURCE_TAIL_POSITION` (oldest transaction not yet committed). If the tail is past the fenced source's final coordinate, everything before it is committed on the target, which is what the stop-based proof establishes, but as a `SELECT`.

Enables `StatusTableEnabled` for new tasks and merges it into the effective settings in `harden` for already-provisioned ones, reusing the stop/modify/resume that phase already performs rather than costing a second interruption.

**Deliberately not wired into the fence.** A new read-only `proof-observe` phase prints the source coordinate, the DMS checkpoint and the target status row side by side with the status row's staleness. Two things must be measured during soak first: the row's update cadence, which is a hard floor on any proof built on it, and whether `SOURCE_TAIL_POSITION` means what the docs say on this engine pair. Read-versus-applied is exactly the confusion that made the old running-task checkpoint assertion wrong; repeating it with a different column would pass a fence over unapplied rows.

The control schema is already excluded from every mapping and object step (`awsdms%` is filtered), so it cannot pollute the migrated set.

## 3. Bounded fence-apply wait, corrected epoch deadline comment

The poll for `read_only=1` was the one unbounded wait left in the phase, out of step with a file where every other wait dies rather than hang. A silently failed or forgotten Spacelift apply extended the outage with no limit.

The source is still writable during that poll, so it is not `read_only` time, but it is customer downtime given the phase's precondition that writers are already scaled to zero. Dying is safe both ways: if the apply never landed nothing changed, and if it landed just after the deadline, re-running `fence` takes the existing re-entry path and reuses the marker. 30m budget, `FENCE_APPLY_TIMEOUT_MIN` overrides.

Also corrects the validation-epoch deadline comment. It claimed real time is roughly `VWAIT x2`, but each loop costs the ~40s paginated describe **plus** the 20s sleep, so the conversion is ~60s per 20 units, i.e. **x3**. At 24k tables that is an ~11.6h ceiling rather than ~7.7h. The comment now states plainly that this bounds a hang and is not an expected duration, since real validation time tracks DMS row-compare throughput rather than table count. Measured epochs recorded for scale: ~25m on a completed mid-size cutover, ~1h for 10.2k/11.9k tables during the gca rehearsal.

## Testing

- `bash -n` and shellcheck `-S warning` clean.
- `status` run against a live in-migration source: reports `fence pre-flight: clear` and the control-table state.
- `proof-observe` correctly refuses when the status table is off.
- Clear, blocking and `FENCE_PREFLIGHT=warn` paths exercised against live AWS state.
- 22 boundary cases on the backup-window comparison including midnight-wrapping windows.
- `checkpoint_coord` verified against the live `RecoveryCheckpoint` value.
- Deadline arithmetic tested including unset and empty-string overrides.

Every call in the new code is a `describe` or a `SELECT`. No production state was changed.

### DIFF
--- a/terraform/physical/static/aurora_migration_settings.json
+++ b/terraform/physical/static/aurora_migration_settings.json
@@ -33,5 +33,8 @@
     "ApplyErrorEscalationPolicy": "STOP_TASK"
   },
   "Logging": { "EnableLogging": true },
-  "ControlTablesSettings": { "ControlSchema": "awsdms_control" }
+  "ControlTablesSettings": {
+    "ControlSchema": "awsdms_control",
+    "StatusTableEnabled": true
+  }
 }
--- a/terraform/physical/static/aurora_migration_settings.json
+++ b/terraform/physical/static/aurora_migration_settings.json
@@ -33,8 +33,5 @@
     "ApplyErrorEscalationPolicy": "STOP_TASK"
   },
   "Logging": { "EnableLogging": true },
-  "ControlTablesSettings": {
-    "ControlSchema": "awsdms_control",
-    "StatusTableEnabled": true
-  }
+  "ControlTablesSettings": { "ControlSchema": "awsdms_control" }
 }
--- a/terraform/physical/util/aurora-migration-runner.sh
+++ b/terraform/physical/util/aurora-migration-runner.sh
@@ -358,6 +358,32 @@ fence_blockers() {
     fi
   fi
 
+  # Validation states, checked HERE rather than only deep inside the fence.
+  #
+  # The fence has TWO validation gates and both fire late: the epoch loop dies on
+  # any state matching mismatch|error|suspend, and the whitelist gate dies on
+  # anything outside {validated, no primary key}. Both run AFTER the source is
+  # fenced, after the drain proof, and after a validation epoch measured in tens
+  # of minutes to hours. So an unclean table today means: spend the entire
+  # customer outage, then refuse.
+  #
+  # Surfacing it in pre-flight turns that into a refusal in seconds at zero
+  # customer cost, and weakens nothing - the late gates still stand. Measured on
+  # a live migration: 3 tables sat in "Table error" the whole time, so this is
+  # the current real state of things, not a hypothetical.
+  #
+  # Skipped when no migration task exists, so 'status' on an unmigrated env stays
+  # quiet. Fails closed on a probe error, like every other check here.
+  TA=$(task_arn 2>/dev/null) || TA=""
+  if [ -n "$TA" ] && [ "$TA" != None ]; then
+    UNC=$(unclean_validation_count 2>/dev/null) || UNC=PROBE_FAILED
+    case "$UNC" in
+      0) : ;;
+      '' | *[!0-9]*) echo "validation-probe: FAILED (could not read the task's validation states - treat as unsafe; got '$UNC')" ;;
+      *) echo "validation-states: $UNC table(s) outside {validated, no primary key}. BOTH fence validation gates will die on these, but only after the source is fenced and the epoch has run. Resolve or prove them before fencing." ;;
+    esac
+  fi
+
   # Match 'creating' ONLY. Negating 'available' would let a snapshot that failed
   # months ago (Status=failed, kept until deleted) block every fence forever, and
   # 'copying' is snapshot-to-snapshot work that does not touch the live source.
--- a/terraform/physical/util/aurora-migration-runner.sh
+++ b/terraform/physical/util/aurora-migration-runner.sh
@@ -411,6 +411,42 @@ timeout_min() {
   printf '%s' "$v"
 }
 
+# Validation PartitionSize for the fence epoch, as a base-10 integer.
+#
+# Why this is pinned rather than left at the AWS default: DMS validation builds a
+# comparison predicate per partition BOUNDARY, and emits malformed SQL (error
+# 1064, an empty predicate term) when a boundary lands on a row whose key column
+# holds an empty string. Such a table reports "Table error" with zero failed and
+# zero suspended records, and the epoch loop dies on it - after the source is
+# already fenced. A table small enough to be a SINGLE partition has no boundary.
+#
+# Measured on a 24,285-table env: 5,205 tables carry a string-typed column in the
+# primary key (any ordinal - the observed failure is on the THIRD key column), and
+# the largest of them holds 425,119 rows by exact COUNT(*). At 1,000,000 every one
+# of them is single-partition, with 2.35x margin.
+#
+# The bounds matter as much as the default. Below 500,000 an at-risk table starts
+# partitioning again and the defect returns. Above 5,000,000 partitions get large
+# enough to drive the replication instance's memory down hard (measured: a
+# 30,000,000 setting drove freeable memory 12.9 -> 5.9 GB and had to be killed).
+# An operator reaching for this knob is by definition inside a customer outage, so
+# the range is enforced rather than documented.
+#
+# Raising it does NOT weaken the proof: a planted single-row divergence was
+# detected identically at 10,000 and at 1,000,000 (Mismatched records, failed=1),
+# on a single-partition at-risk table, with row counts equal on both sides so a
+# count check alone would have missed it.
+validation_partition_size() {
+  local v="${FENCE_VALIDATION_PARTITION_SIZE:-1000000}"
+  case "$v" in ''|*[!0-9]*) die "FENCE_VALIDATION_PARTITION_SIZE must be a positive integer (got '$v')" ;; esac
+  v=$((10#$v))
+  { [ "$v" -ge 500000 ] && [ "$v" -le 5000000 ]; } || \
+    die "FENCE_VALIDATION_PARTITION_SIZE must be between 500000 and 5000000 (got $v). Below 500000 re-partitions
+the at-risk tables and reintroduces the empty-string-key defect; above 5000000 moves toward the partition sizes
+that exhausted the replication instance's memory."
+  printf '%s' "$v"
+}
+
 # $1 = a short label for the log line, so the two call sites are distinguishable
 preflight_fence_gate() {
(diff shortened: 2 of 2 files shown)


ddvdozuki:
Two more commits, both from an audit of the fence phase's timing.

**`f17feb7` instrumentation.** Nothing has ever measured where the fence's time actually goes, so every figure in planning is an estimate. The validation epoch now logs per-loop wall clock split into API time versus sleep, tables cleared since the last sample, and a total at the end. That settles three open questions at once: whether the VWAIT-to-wall factor is 2 or 3, whether the loop is API-bound or validation-bound, and whether convergence tracks table count or data volume. The deadline formula assumes count, but row-compare is a volume operation and that assumption has never been checked.

Also adds freeze-envelope milestones and a note at GO that the freeze does **not** end there: the source stays fenced through the physical apply, its auto-following logical apply and the pod rollout. That tail is untimed and may exceed everything measured above.

**`f17feb7` also rewrites the writer precondition.** It claimed "app writers scaled to zero" while the kill sweep's own comment fifty lines later assumes a live app whose pool reconnects between sweeps. Both modes work (the cutoff is `read_only=1` either way) but the downtime profiles differ completely: drained, the outage started before this phase; live, every gate including the multi-hour epoch is customer-visible write downtime.

**`6e81124`** gives `preload` the 21600s SSM budget the other heavy DDL steps already pass. It dumps every schema's DDL (24k+ tables, one `SHOW CREATE TABLE` each) on the 3600 default.

### One recommendation from the audit I rejected

Moving `vtask_teardown` past the GO line, for a bounded 1500s saving. The validation task is created against the **same** source and target endpoint ARNs that the cutover apply modifies, and this runner already stops the BI task ahead of the repoint for precisely that reason ("DMS rejects endpoint modification on a running task"). The operator acts on GO immediately, so moving teardown after it races the endpoint repoint. The current ordering is deliberate.

### Numbers corrected while doing this

- Drain-proof worst case is **3.46h** (10 cycles x (300+45+900)), not the 5.96h I first wrote, which wrongly charged the one-time initial stop to every cycle.
- `VMAX` is an **abort threshold**, not a duration. Hitting it dies with the source fenced, the main task stopped and a validation task orphaned. Neither 7.7h nor the corrected 11.6h should be quoted as a window.

ddvdozuki:
### Control table dropped

Reverted `a2f0c54`. It was speculative and the cost/benefit does not hold up:

- It buys one thing, replacing the drain proof's DMS task stop with a `SELECT`. That stop is bounded at 900s and typically minutes, against a validation epoch measured at 25 min to hours. Optimising the small term again.
- It cannot be wired in until soak measures the update cadence and the read-vs-applied semantics of `SOURCE_TAIL_POSITION`, so it does nothing for the migration in flight either way.
- Enabling it on an existing task costs a stop/resume of live CDC, which is the operation it exists to avoid.
- The `harden` change turned a previously no-op `harden` into a task stop for any task lacking the table, a real regression on a mid-migration environment, introduced purely to enable something nothing needs yet.

Worth revisiting if the fast drain proof is ever actually pursued, on a newly provisioned environment rather than one mid-cutover.

### Review defects fixed (`274c40c`)

**The gate failed open on 3 of 5 probes.** It is consumed as `blockers=$(fence_blockers)`, and errexit is not in force inside a command substitution used as an assignment RHS, so a probe that died contributed nothing and an empty result read as "clear". Confirmed on bash 3.2 and 5.2.

Not hypothetical: `rds:DescribeDBSnapshots` is a permission this gate introduced, so a runner role predating it would have silently reported "no snapshot running" on every fence forever, defeating the check whose whole point is the safety-snapshot case. Every probe now resolves to an explicit sentinel that emits a blocker.

```
all probes healthy                         *** PASSED (would fence) ***
rds:DescribeDBSnapshots DENIED             BLOCKED:
      - snapshot-probe: FAILED (describe-db-snapshots unreadable - treat as unsafe...)
rds:DescribeDBInstances DENIED             BLOCKED:
      - instance-probe: FAILED (describe-db-instances unreadable - treat as unsafe...)
```

**The entry gate stood in front of the fenced-source recovery path.** It ran before the `RO0` already-fenced probe, so re-running `fence` after a mid-fence death (source read_only, customers down) hit blockers that are meaningless on re-entry. Moved into the fresh-fence branch.

**Snapshot matching named the wrong states.** `Status!='available'` catches a snapshot that failed months ago and would block every fence forever. Now matches `creating`/`copying`/`modifying` explicitly.

Also folded four `describe-db-instances` calls into one (same response, four throttle/permission exposures), made the window helpers pure functions over the window string, and bounded `abort`'s poll for `read_only=0` on the same budget. That last one is the path taken when things have already gone wrong, so hanging forever was the worst remaining failure mode.

**Verification:** 22-case window boundary suite still passes including midnight wrap; malformed window strings return a blocker rather than crashing or reporting OPEN; snapshot-state filter checked against all five states; live `status` still reports `fence pre-flight: clear`.

ddvdozuki:
## Ready for review

Context has changed since this PR was opened, so worth stating plainly what it is now.

**No environment currently uses this path.** Every env that migrated with this runner has completed and sits on `aurora_migration_state = "off"`. The last candidate abandoned the DMS route rather than finishing it, on measured grounds recorded in the runner header.

**So why merge it?** The runner ships in this repo either way. Closing this PR does not remove the buggy version, it just leaves it in place. The worst bug it fixes is real:

> The epoch loop polled only table statistics. A validation task that failed, was deleted, or exhausted the replication instance's memory left the stats stale and every gate reading clean-but-incomplete (no bad states, no enumeration drift, nothing ever clearing). The loop then ground all the way to `VMAX` **with the source `read_only`** - about **11.6 hours** at 24k tables. `VMAX` is documented as a bound on a hang; it was the only bound on this one.

That is exactly the class of latent bug that survives because nobody expects to hit it.

**Merging cannot affect any running stack.** This PR touches one file, and that file is not referenced by any Terraform code anywhere in the repo - it is a manually invoked utility (`aurora-migration-runner.sh <identifier> <region> <phase>`). No apply executes it.

### What is in it

- **Fence pre-flight gate** on backups, snapshots and in-flight parameter applies. A deferred apply was measured at 262.89s against a normal 3.3-6.7s, and the unfence is REFUSED for that whole window, so the abort path is unavailable while it lasts.
- **A read-only `writers` phase** that measures whether app writes actually drained, using DMS per-table counters rather than server-wide ones. Measured on a live source: `Handler_write` 4099/s and `Innodb_rows_inserted` 9.2/s against a real business write rate of 0.7/s, so the server-wide counters would tell an operator the drain never worked.
- **`PartitionSize` pinned at 1,000,000**, bounded 500k-5M. DMS emits malformed SQL (error 1064, an empty predicate term) when a validation partition boundary lands on a row whose key column holds an empty string; a single-partition table has no boundary. 5,205 tables carry a string-typed PK column (at any ordinal - the observed failure is on the third), largest 425,119 rows by exact `COUNT(*)`.
- **Epoch watchdog**: task-health check each poll plus a freeable-memory floor, so a dead task or an OOM fails loudly instead of hanging to `VMAX`.
- **Post-epoch gate narrowed**: a stale `Table error` is forgiven only when the fresh epoch re-validated that exact table. `Mismatched`, `Suspended`, `Error` and unknown states still refuse. Demotion is by identity and by that one state, never by count.
- **Header note** recording that the DMS route is not the recommended path for a large table count, and warning off the Terraform provider's `blue_green_update` (it switches over unattended and then deletes the old primary with `SkipFinalSnapshot=true`).

### Verification

- 38 tests covering the partition-size bounds, the watchdog state machine, the memory floor comparison and the post-epoch gate. Bound to the real runner functions, not copies.
- Red-green verified by mutating the implementation three ways (forgive any state, drop the `PartitionSize` floor, make a missing metric read as 0) - each caught by the right tests.
- **The correctness claim was measured, not assumed.** A planted single-row divergence on an at-risk single-partition table was detected identically at `PartitionSize` 10,000 and 1,000,000 (`Mismatched records, failed=1`), with row counts equal on both sides so a count check alone would have missed it. Target restored and verified back to its exact pre-test checksum.
- `bash -n` and shellcheck clean.

PR #429, opened 2026-08-02
feat(rds): allow the provisioned primary to change engine version via blue/green

ddvdozuki:
Adds `rds_blue_green_update` (default **false**), wiring the RDS module's `blue_green_update` block.

## Why

For the MySQL 8.0 -> 8.4 major upgrade. 8.0 left standard support on 2026-07-31, so anything still on it bills Extended Support per vCPU-hour.

An **in-place** major upgrade restarts the engine and runs the upgrade on the live instance: minutes of hard downtime on a large multi-tenant stack, with no confident upper bound.

**Blue/green** builds a second environment at the new version, replicates into it, and switches over once caught up. AWS puts the switchover at roughly a minute, and critically it **preserves the endpoint hostname** - so the credentials secret, the DMS endpoints, the bastion association and the app's own config all keep resolving correctly. No repoint, no rollout.

## Why default false

- It runs a full second copy of the database for the duration, so it is not free.
- It has prerequisites a stack can fail silently: binlog on (`backup_retention_period > 0`), and the instance must not itself be a read replica.

Intended use is to turn it on for the apply that changes the version, then turn it off.

## Caveat, documented in the variable

Moving `rds_engine_family` also **replaces the parameter group** (its family derives from that variable), so such a plan couples a blue/green instance update with a parameter-group replacement. That plan should be read, not confirmed blind.

## Verification

- `tofu fmt` clean, tflint clean, terraform-docs regenerated
- `tofu validate` outside terragrunt fails on this root module for a pre-existing reason (missing provider aliases that terragrunt supplies) - confirmed identical on pristine `master`, not introduced here


ddvdozuki:
## On hold - do not merge

Two independent reviews found that the AWS provider's `blue_green_update` path is **more destructive than the AWS console flow**, which makes this the wrong mechanism for a production primary.

Provider 6.56.0, `internal/service/rds/instance.go`, after switchover:

1. moves Terraform state to the green instance's `DbiResourceId`
2. issues `ModifyDBInstance` with `DeletionProtection: false` on the **old blue primary**
3. calls `DeleteDBInstance` with **`SkipFinalSnapshot: true`**
4. waits for the old production instance to disappear

It never sets `DeleteAutomatedBackups`, so the automated backups go too. Manual snapshots survive; the provider takes none.

This bypasses every guard in this repo:

| guard | outcome |
|---|---|
| `deletion_protection = true` (rds.tf:210) | provider explicitly disables it |
| `skip_final_snapshot = false` | ignored on this path |
| `final_snapshot_identifier_prefix` | ignored on this path |
| `db-replace-guard` OPA policy | **cannot fire** - blue/green plans as `changed`, which is safelisted, and the deletion happens inside the provider after the plan policy passed |

It also switches over the moment green reports available, with no pause to validate replica lag or run checks, and the whole operation is bounded by an unconfigured `Update: schema.DefaultTimeout(80 * time.Minute)`.

By contrast AWS's own flow **retains** blue after switchover so it can be inspected or rolled back to.

### What we are doing instead

Driving the blue/green through the CLI, where switchover is a deliberate human step and blue is kept as the rollback. Terraform gets reconciled afterwards by setting `rds_engine_family = "8.4"`.

### Should this still merge eventually?

The wiring itself is correct and both reviewers confirmed the `{}` default is a genuine no-op that perturbs no existing stack. If it is ever revived it needs, at minimum, `timeouts = { update = "4h" }` and a very loud warning in the variable description about the undefended deletion of the old primary. Leaving it in draft rather than closing, since the analysis is worth keeping.

ddvdozuki:
## Closing - the mechanism is unsafe, and merging it would make it discoverable

Not closing because the code is wrong. Both reviewers confirmed the wiring is correct and that the `{}` default is a genuine no-op that perturbs no existing stack. Closing because the capability itself should not be reachable from this module.

### Why

The AWS provider's `blue_green_update` path (6.56.0, `internal/service/rds/instance.go`) does this immediately after switchover, with no human step in between:

1. moves Terraform state to the green instance's `DbiResourceId`
2. issues `ModifyDBInstance` with `DeletionProtection: false` on the **old blue primary**
3. calls `DeleteDBInstance` with **`SkipFinalSnapshot: true`**
4. waits for the old production instance to disappear

It never sets `DeleteAutomatedBackups`, so the automated backups go with it. Manual snapshots survive; the provider takes none.

Every guard in this repo is bypassed:

| guard | outcome |
|---|---|
| `deletion_protection = true` (rds.tf) | provider explicitly disables it |
| `skip_final_snapshot = false` | ignored on this path |
| `final_snapshot_identifier_prefix` | ignored on this path |
| `db-replace-guard` OPA policy | **cannot fire** - blue/green plans as `changed`, which is safelisted, and the deletion happens inside the provider after the plan policy has already passed |

It also switches over the moment green reports `available`, with no pause to check replica lag or validate anything, and the entire operation is bounded by an unconfigured `Update: schema.DefaultTimeout(80 * time.Minute)` - on timeout the green build is discarded.

AWS's own flow **retains** blue after switchover precisely so it can be inspected or rolled back to. The provider is strictly more destructive than the console.

### Why closing rather than parking

The danger lives inside the provider, so nothing in this repo can fix it - only document it. Merging would put a variable called `rds_blue_green_update` into the module, and the next person to find it would reasonably assume it is a safe, supported way to change an engine version. A variable whose correct usage note is "do not use this" should not be discoverable.

### What we do instead

Drive the blue/green from the CLI, where switchover is a deliberate human step and blue is retained as the rollback:

```
create-blue-green-deployment    # green builds and replicates, production untouched
  ...validate green, watch replica lag...
switchover-blue-green-deployment --switchover-timeout N
  ...blue is renamed -oldN and RETAINED, read-only...
```

Then reconcile Terraform by setting `rds_engine_family` to the new version so the next plan converges.

Rollback boundaries under that flow, from the AWS docs:
- **before switchover**: fully reversible, delete the deployment and blue is untouched
- **during switchover**: atomic - *"if the switchover starts and then stops before finishing for any reason, then any changes are rolled back, and no changes are made to either environment"*
- **after switchover**: blue is retained but frozen read-only and replication stops, so a rollback loses writes made to green
- **after deleting blue**: final point of no return

### Where the analysis lives

The header note merged in #426 records this warning next to the migration runner, so it is in the repo rather than only in a closed PR.

PR #431, opened 2026-08-03
fix(physical): set final_snapshot_identifier on both aurora clusters

ddvdozuki:
Found while aborting an Aurora migration: with `protect_resources = true`, neither
Aurora cluster can be destroyed through terraform at all.

## The bug

`module.aurora` (aurora.tf) and `module.bi_aurora` (bi.tf) both set:

```hcl
deletion_protection = var.protect_resources
skip_final_snapshot = !var.protect_resources
```

but neither sets `final_snapshot_identifier`, and rds-aurora v10.2.0 defaults it to
`null`. The provider hard-errors on delete when a final snapshot is required and no
name was given.

The rds and provisioned-BI paths already avoid this with
`final_snapshot_identifier_prefix` (rds.tf:208, bi.tf:575), which the rds module
suffixes with a `random_id`. rds-aurora exposes no prefix variable, only the literal
identifier, so the suffix is built here from one shared `random_id` covering both
clusters.

## Why it matters

It breaks the `aurora_migration_state = "off"` abort path specifically. Deletion
protection is only the first gate, and it can be cleared out of band. The missing
snapshot name is the second, and it cannot: `skip_final_snapshot` is a terraform-only
attribute with no RDS API equivalent, so there is nothing to flip with the CLI. The
cluster has to be snapshotted and deleted by hand, then dropped from state by refresh.

That is the documented reversal for a migration that gets provisioned and then
abandoned, so it should work.

## Scope and blast radius

Adding `final_snapshot_identifier` shows as an in-place update on existing clusters
and issues no API call, since the attribute is only read at destroy time. Both new
`random_id` and the attribute are gated so they exist exactly when their cluster does:

```hcl
count = local.aurora_present || local.bi_uses_aurora ? 1 : 0
```

`local.aurora_present` covers `module.aurora`, `local.bi_uses_aurora` covers
`module.bi_aurora`, so the `[0]` index always resolves.

Note `aws_rds_cluster` is in auto-gate's `sensitive_types`, so envs picking this up
will gate for a manual confirm rather than autodeploying. That is the intended
behaviour for a DB-touching diff.

## Known limitation

A destroy does not consume the `random_id`, so a stack destroyed, recreated and
destroyed again reuses the hex and the second delete fails `DBSnapshotAlreadyExists`.
That is the same behaviour as the existing rds and BI paths rather than a regression;
rename or drop the stale snapshot when it happens.

## Verification

`tofu fmt` and the pre-commit `tflint` pass. `tofu validate` fails on this module
both with and without the change, identically, because it expects the caller to
configure the `aws.dr` provider alias that terragrunt generates. Confirmed by
validating pristine `master` in a separate worktree.

### DIFF
--- a/terraform/physical/README.md
+++ b/terraform/physical/README.md
@@ -198,6 +198,7 @@
 | [null_resource.create_dms_cloudwatch_role](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
 | [null_resource.create_dms_vpc_role](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
 | [null_resource.s3_replication_job_init](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
+| [random_id.aurora_final_snapshot](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/id) | resource |
 | [random_password.aurora](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password) | resource |
 | [random_password.bi_aurora](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password) | resource |
 | [terraform_data.aurora_migration_phase](https://registry.terraform.io/providers/hashicorp/terraform/latest/docs/resources/data) | resource |
--- a/terraform/physical/aurora.tf
+++ b/terraform/physical/aurora.tf
@@ -16,6 +16,20 @@ resource "random_password" "aurora" {
   special = false
 }
 
+# Suffix for the Aurora final-snapshot names (this file and bi.tf). The rds-aurora
+# module has no final_snapshot_identifier_PREFIX variable the way the rds module
+# does, only the literal identifier, so the uniqueness the rds path gets for free
+# has to be built here. Same shape as what terraform-aws-modules/rds appends.
+#
+# A destroy does not consume the value, so a stack that is destroyed, recreated
+# and destroyed again reuses this hex and the second delete fails
+# DBSnapshotAlreadyExists. That matches the existing rds and BI paths rather than
+# regressing them; rename or drop the stale snapshot when it happens.
+resource "random_id" "aurora_final_snapshot" {
+  count       = local.aurora_present || local.bi_uses_aurora ? 1 : 0
+  byte_length = 4
+}
+
 module "aurora" {
   source  = "terraform-aws-modules/rds-aurora/aws"
   version = "10.2.0"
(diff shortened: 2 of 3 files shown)


ddvdozuki:
Bumping priority slightly: usac applied v8.10.0 tonight, which creates `module.bi_aurora`. That cluster is therefore born with `deletion_protection = true`, `skip_final_snapshot = false` and no `final_snapshot_identifier`, so it cannot currently be destroyed through terraform.

Not urgent and nothing is broken, but it stops being hypothetical: if the BI full load needs a clean target, tearing that cluster down means the same out-of-band snapshot-and-delete this PR removes the need for.

Landing it later is cheap. `final_snapshot_identifier` is not ForceNew on `aws_rds_cluster` (read only at delete time), so it applies as an in-place update on existing clusters, no replacement.

ddvdozuki:
Added a second commit after tonight's apply hit the other half of this problem.

## `bi_treat_as_derived`, default true

BI is a DMS target. It holds nothing the primary does not, and a full load rebuilds it, which is exactly what the rds -> aurora switch does on purpose. `module.rds_replica_database` (the legacy read-replica BI path, same file) has always set `deletion_protection = false` and `skip_final_snapshot = true` for that reason. The two DMS paths were the inconsistent ones, carrying `protect_resources` through to a database nobody would ever restore.

Both ways that bit tonight on the same apply:

**1. The final snapshot outran the credentials.** usac's BI instance is 115 GB. Its final snapshot took over an hour and `aws_db_instance` destroy waits on it. Spacelift's AWS session is capped at **1 hour by role chaining** (`spacelift-deployer` then the member account's `OrganizationAccountAccessRole`), and that cap cannot be raised on either role, it is an AWS limit on chained sessions rather than a setting. So a snapshot nobody wanted can outlive the session and fail the apply.

**2. Deletion protection needed an out-of-band pre-step.** The `bi_db_engine` docs already spell this out at length: clear it by hand first or the apply dies mid-way, after the Aurora cluster exists and with replication stopped against the 48h deprovision clock. That whole paragraph exists only because BI was treated as precious.

## Default choice, flagging explicitly

I set the default to **true** (derived) rather than preserving current behaviour, because:

- it matches what the read-replica path in the same file already does
- `skip_final_snapshot` and `deletion_protection` are not ForceNew, so this is an in-place update on existing stacks, no replacement and no plan churn beyond the attribute itself
- it only changes destroy-time behaviour, and the destroy it changes is one the docs already describe as safe to redo

Say if you would rather it defaulted false and each env opted in. `final_snapshot_identifier` stays set either way so flipping to false still works.

Scoped to the DMS target only. The primary's `protect_resources` is untouched.

ddvdozuki:
Correcting my own noise on this: I retitled this PR and posted a `bi_treat_as_derived` comment above **after it had already merged**, so both were misleading. The title is restored to what actually shipped in 8.11.0, which is the `final_snapshot_identifier` fix only.

The BI derived-treatment change is now #434, rebased onto master on top of this one.

PR #453, opened 2026-08-05
chore(tofu): bump CPI harness tofu pin to 1.11.13 (phase 1b)

ddvdozuki:
**MUST merge only after the infra-live fleet bump lands** (`_spacelift/stacks.tf` `terraform_version` `1.11.8` -> `1.11.13`, terragrunt untouched at `0.99.4`). This PR just moves the CPI-side harness/CI pins to match; merging it first would make the harness image test an engine Spacelift isn't running yet. Draft until that infra-live change is in.

## What

Phase 1b of the tofu/terragrunt upgrade study: tofu 1.11.8 -> 1.11.13, terragrunt held at 0.99.4 everywhere (no terragrunt change - 1.1.x still double-assumes `iam_role`).

Pin sites bumped:
- `live/tests/Dockerfile` (`ARG TOFU_VERSION`)
- `live/tests/run.sh` (`FLEET_TOFU_VERSION`, the local non-container drift warning)
- `live/Dockerfile` (the separate `live/` dev-harness image)
- `azure-config/.opentofu-version` - single source of truth for three CI consumers (`terraform.yml`, `provider-lock.yml`, `release-azure.yml`), all read from this file so no separate workflow edits needed there.

Also updated three comments that named `1.11.8` literally and would otherwise go stale: `terraform/physical-azure/main.tf`, `azure-config/README.md`, `.github/workflows/terraform.yml`. Left the `CHANGELOG.md` entry and the `live/tests/run.sh` anecdote about a past 1.12.2 run alone - both describe history, not the current pin.

`required_version` floors are untouched - `>= 1.11.1` (physical/physical-azure) and `>= 1.9.0` (logical) are both satisfied by 1.11.13 same as 1.11.8.

## Why 1.11.13 and not 1.12.5

1.11.13 clears the same 17 CVE findings as 1.12.5 (net -16, landing at 70 from the phase-1 baseline of 86) while staying inside the terragrunt 0.99.4-supported matrix and touching neither the terragrunt regression nor the 1.12 `replace_triggered_by` change. It also patches `GHSA-q7j3-v8qv-22vq` outright - that's fixed in the 1.11 line at 1.11.10, not only at 1.12.3. Tradeoff: 1.11.x left support Aug 1, so 1.11.13 is terminal for that line.

## Sequencing

1. Land the infra-live `stacks.tf` bump first (canary `mpc-dev-min`, per the upgrade study - no separate canary needed for 1b since it makes no terragrunt change and doesn't touch the credential-chain regression that phase 2 needs a different canary for).
2. Mark this PR ready and merge once the fleet is actually on 1.11.13, so the harness never diverges from the engine Spacelift runs.


ddvdozuki:
On the rebase after #452 merges: also bump DEPLOYMENT.md's OpenTofu row (1.11.8 -> 1.11.13) so the deploy doc tracks the fleet pin this PR mirrors. Flagged in #452's review as the cross-PR staleness gap.

PR #456, opened 2026-08-05
chore(master): release 8.15.0

dozuki-machine-user:
:robot: I have created a release *beep* *boop*
---

### Features

* **logical:** plumb the Watchdog heartbeat flag through to the chart ([c62cc15](https://github.com/Dozuki/CloudPrem-Infra/commit/c62cc1550cb64adc58197331353c027dec0d4d92))
* **tofu:** bump CPI harness/CI pins to phase 2 (tofu 1.12.5, terragrunt 1.1.2) ([#455](https://github.com/Dozuki/CloudPrem-Infra/issues/455)) ([514b6b6](https://github.com/Dozuki/CloudPrem-Infra/commit/514b6b609dbc5b48df9ca286fdfd12ab7b03d934))

### Bug Fixes

* **logical:** set grafana root_url so the dashboards UI loads under /grafana ([bd08d51](https://github.com/Dozuki/CloudPrem-Infra/commit/bd08d51950359fa3596293f2e5b74c3af174c008))
* **logical:** stop istiod replicas colocating on one node ([#454](https://github.com/Dozuki/CloudPrem-Infra/issues/454)) ([9da3f0c](https://github.com/Dozuki/CloudPrem-Infra/commit/9da3f0ca2be3cd2883a16bf2c3e825a91514e925))
* **physical:** DR bucket backfill on creation, replication metrics + alarm, DR-region access logs ([#458](https://github.com/Dozuki/CloudPrem-Infra/issues/458)) ([7b2eb40](https://github.com/Dozuki/CloudPrem-Infra/commit/7b2eb40d2459dbbce66ba5373572cfc03504d29c))
* **physical:** gate the dms event rule on dms_enabled so pure-replication BI can apply ([8407dcf](https://github.com/Dozuki/CloudPrem-Infra/commit/8407dcfa2f47adaf6db28ad124524ca8ddd7f277))
* **physical:** widen BI CDC latency alarm window to 9-of-12 ([#457](https://github.com/Dozuki/CloudPrem-Infra/issues/457)) ([d401227](https://github.com/Dozuki/CloudPrem-Infra/commit/d4012276e4c887938c651134085ebab6ca12c593))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #459, opened 2026-08-05
test(harness): stop superseded matrix runs for the same PR at submission

ddvdozuki:
Seen live on the 8.15.0 release PR (#456): three master pushes in one afternoon left four matrix runs for the same PR, with both harness semaphore slots held by children testing superseded shas while the real head's children sat queued. The staleness gate only fires at child admission, so runs already past it burned slots for hours.

Supersession runs in the GitHub submit job, BEFORE the new matrix is created: stop every matrix carrying this PR's harness/pr label (children swept only under parents whose patch succeeded), then create the new run. Running it pre-creation is what makes it safe - everything selected is older by construction, so there is no self-exclusion, no name parsing, and no timestamp tiebreak. Stop rather than terminate so the scenario teardown exit handler still runs, and the workflow-level semaphore is held through that teardown, so a superseded stack is always fully destroyed before a waiting child takes its slot. The harness/superseded marker rides the same patch as the stop; the matrix's notify handler suppresses the commit status, PR comment and Slack post for marked runs (retried lookup, fail-open, re-checked after evidence collection).

Security note: the patch verb lands ONLY on harness-pr-submitter, the Role bound to the GitHub OIDC identity - the shared argo-workflow SA (Terraform phase pods included) gains nothing. An earlier revision granted patch to the shared SA; the panel review rejected that, and an invariant test now locks it out.

Rollout notes:
- Matrices submitted before this merges lack the harness/pr label and cannot be superseded; they drain naturally (one-time gap, accepted).
- Superseded shas keep their pending commit status forever; cosmetic, branch protection evaluates the current head.
- After merge the templates need a hand apply per live/tests/argo/README.md: kubectl --context argo-ops -n argo apply -f live/tests/argo/

Reviewed by a 4-seat consensus panel plus the senior reviewer across three rounds; findings and adjudications are in the review thread.

### DIFF
--- a/.github/workflows/upgrade-tests.yml
+++ b/.github/workflows/upgrade-tests.yml
@@ -49,6 +49,10 @@ env:
   PR_REPO: ${{ github.event_name == 'pull_request' && github.repository || '' }}
   PR_SHA: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || '' }}
   PR_BASE_REF: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.ref || '' }}
+  # Supersession identity for the matrix's supersede step: it selects older runs by
+  # this label, never by parsing workflow names. Empty on dispatch runs, which never
+  # supersede anything.
+  PR_NUM: ${{ github.event_name == 'pull_request' && github.event.pull_request.number || '' }}
   NAME_HINT: ${{ github.event_name == 'pull_request' && format('pr{0}', github.event.pull_request.number) || 'manual' }}
 
 jobs:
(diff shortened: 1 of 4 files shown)


ddvdozuki:
Multi-model consensus review ran on de1e1bc: 7 reviewers, verdict request-changes with 3 majors (pod-level failures could block the fan-out and post a false failure to the head sha; superseded runs posted red status/comment/Slack noise on every force-push; child-stop ordering had a race that could orphan a late-created child) and 4 minors (identity by name parsing, shared SA gaining the patch verb, tiebreak semantics, missing invariant coverage).

6a30963 addresses all of it: continueOn failed+error plus 10s kubectl request timeouts, an atomic superseded-marker label that post-status suppresses on, parent-first stop with a bounded retrying child sweep across parents in any phase, identity via a harness/pr label set at submission (name parsing removed, params env-passed), a dedicated harness-matrix ServiceAccount for the patch verb, and invariant tests pinning every property above. Timestamp ties are deliberately left alone: both proposed tiebreakers were shown able to stop the newer run.

After merge the templates need the hand apply per live/tests/argo/README.md.

PR #467, opened 2026-08-08
fix(dms): set an explicit timeout on the dms_restart lambda

ddvdozuki:
The `dms_restart` Lambda never set a `timeout`, so it ran on AWS's 3-second default while doing a `FilterLogEvents` plus a `StartReplication` on a cold arm64 start.

That is not theoretical headroom. Real invocations land at 2765-2931 ms, and one has already been killed outright:

```
REPORT RequestId: ...
Duration: 3000.00 ms
Billed Duration: 3000 ms
Memory Size: 128 MB
Status: timeout
```

It recovered only because SNS retried the invocation.

A restart lost this way is silent. The replication stays failed against the 48h deprovision deadline, after which recovery is no longer a start call but a Terraform apply, and the only signal is the CRITICAL state-change alert that had already fired for the original failure.

`30` matches `sns_to_slack` in the same file, which has always set it. This one looks like a simple omission. Lambda bills the duration actually used, so the higher ceiling costs nothing on the normal ~2s path.

## Scope

This PR originally carried a `resume-processing` before `reload-target` recovery ladder. Multi-model review found four MAJOR defects in it, including a suppression-marker bug that produced an alternating `reload -> suppress -> reload` loop, and a conflict with the documented invariant at `bi.tf:197` that a repointed (empty) target must be reloaded rather than resumed. That work is dropped; the findings are preserved in the review comment below.

The root cause it was chasing is better fixed at the DMS layer: `RecoverableErrorCount` is pinned to `0` in `static/dms_config.json`, which disables DMS's own retry of a lost target connection. AWS's default is `-1` (nine retries for exactly that failure). That change is tracked separately.

What remains here is the one independently verified, self-contained bug.

### DIFF
--- a/terraform/physical/monitoring.tf
+++ b/terraform/physical/monitoring.tf
@@ -833,6 +833,16 @@ resource "aws_lambda_function" "dms_restart" {
   runtime       = "python3.12"
   architectures = ["arm64"]
   role          = aws_iam_role.lambda_execution[0].arn
+  # This was never set, so the function ran on AWS's 3s default while doing a
+  # FilterLogEvents plus a StartReplication on a cold arm64 start. Real invocations
+  # land at 2.7-2.9s and one has already been killed outright ("Status: timeout",
+  # 3000.00 ms), recovered only because Lambda retried the async invocation. That
+  # rescue is not guaranteed: once the async retries are spent, a restart lost this
+  # way is silent, and the replication sits failed against the 48h deprovision clock
+  # with nothing but the CRITICAL state-change alert to say so. 30s matches
+  # sns_to_slack above; Lambda bills duration actually used, so a higher ceiling
+  # costs nothing on the normal 2s path.
+  timeout = 30
 
   source_code_hash = data.archive_file.dms_restart_lambda[0].output_base64sha256
 


ddvdozuki:
Consensus review: **REQUEST CHANGES**. Three reviewers, all independently "not ready to merge". Four MAJOR findings, three of which I reproduced directly. Recording them here; the branch is staying draft.

### 1. The suppression line defeats the ladder and causes an alternating reload loop (CRITICAL)

The fast-fail suppression message itself contains `AUTO-RESTART` and the ARN, but no strategy token. It therefore matches the CloudWatch filter, matches the ARN filter, and parses as `resume-processing`. Reproduced against the real emitted string:

```
failure 1: newest marker = reload-target      -> SUPPRESS
failure 2: newest marker = resume-processing  -> START reload-target
failure 3: newest marker = reload-target      -> SUPPRESS
failure 4: newest marker = resume-processing  -> START reload-target
```

The old code was self-reinforcing here (any marker meant suppress). This change converts that into `reload -> suppress -> reload -> suppress`, reintroducing the production-reload loop the cooldown exists to prevent. Same root cause makes a legacy `AUTO-RESTART <arn>` marker parse as a resume when it actually represented a reload.

### 2. Resume-first is unsafe across a target repoint (MAJOR)

`bi.tf:197` and `variables.tf:356` already document this: the RDS to Aurora BI switch modifies the target endpoint **in place**, so neither the endpoint ARN nor the replication-config ARN changes. A delayed or duplicate failure event after the repoint would now resume CDC against the empty cluster and report `Running` with no historical rows. The previous always-reload behaviour was safe here by construction.

### 3. Bare `except Exception` escalates to a destructive reload on any start failure (MAJOR)

`InvalidResourceStateFault` from a replication still mid-transition is the common case, and it is not evidence of an unusable checkpoint. Confirmed by hand: a start against a settling replication returns `... is in MODIFYING state and cannot start`. Under this code that discards the target and re-reads every table from the production source.

### 4. Marker records intent, not outcome (MAJOR)

The marker is printed before the start call, so a rejected, timed-out, or never-executed call is recorded as though that rung happened. Combined with SNS at-least-once delivery, two concurrent invocations can both read no marker, both resume, and the loser records a reload it never performed.

### Separate, pre-existing, and independently verified: the function has no timeout set

`aws_lambda_function.dms_restart` sets neither `timeout` nor `memory_size`, so it runs on the AWS default of 3s/128MB. Observed durations are 2765-2931 ms and one production invocation was killed outright:

```
REPORT RequestId: ...
Duration: 3000.00 ms
Status: timeout
```

That is a real bug independent of this PR, and it makes any added API call in the restart path unsafe. Worth fixing on its own.

### Also raised

- `filter_log_events` pagination is unhandled; a truncated first page can hide the newest marker (default order is ascending).
- The 30-minute window means failures spaced further apart always resume and never escalate.
- No test harness exists for this Lambda.

### Where this leaves the change

The root-cause fix for the failure that motivated this PR is `RecoverableErrorCount: -1` (AWS default; this repo pins it to `0`, which disables DMS internal retry entirely). With that set, DMS retries a lost target connection itself and the replication never reaches `failed`, so this Lambda never fires for the case the ladder was built to optimise. That materially shrinks the value of resume-first while the hazards above remain, so the ladder needs either a real state store or a narrower trigger rather than a patch-up.

